PR #16, opened 2015-03-16
Added fedmsg messages to mirrormanager2

Devyani-Divs:
@ralphbean @pypingou Messages are sent on fedmsg when a mirror is added and when a mirror is updated as discussed in issue fedora-infra/fedmsg#245 . Please review. Thankyou :)

### DIFF
--- a/fedmsg.d/example-config.py
+++ b/fedmsg.d/example-config.py
@@ -1,5 +1,5 @@
 import socket
-hostname = socket.gethostname()
+hostname = socket.gethostname().split('.')[0]
 
 config = dict(
     endpoints={
--- a/mirrormanager2/app.py
+++ b/mirrormanager2/app.py
@@ -351,12 +351,14 @@ def site_view(site_id):
         form=form,
     )
 
-
+from lib.notifications import fedmsg_publish
 @APP.route('/host/<int:site_id>/new', methods=['GET', 'POST'])
 @login_required
 def host_new(site_id):
     """ Create a new host.
     """
+
+    topic = 'host.added'
     siteobj = mmlib.get_site(SESSION, site_id)
 
     if siteobj is None:
@@ -370,10 +372,15 @@ def host_new(site_id):
         form.populate_obj(obj=host)
         host.bandwidth_int = int(host.bandwidth_int)
         host.asn = None if not host.asn else int(host.asn)
+        message = dict(
+            site_id=host.site_id,
+            bandwidth=host.bandwidth_int,
+            asn=host.asn)
 
         try:
             SESSION.flush()
             flask.flash('Host added')
+            fedmsg_publish(topic, message)
         except SQLAlchemyError as err:
             SESSION.rollback()
             flask.flash('Could not create the new host')
@@ -477,6 +484,7 @@ def siteadmin_delete(site_id, admin_id):
 def host_view(host_id):
     """ Create a new host.
     """
+    topic = 'host.updated'
     hostobj = mmlib.get_host(SESSION, host_id)
 
     if hostobj is None:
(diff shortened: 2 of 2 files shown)


fedmsg.d/example-config.py
```diff
@@ -1,5 +1,5 @@
 import socket
-hostname = socket.gethostname()
+hostname = socket.gethostname().split('.')[0]
```
ralphbean: The change on this line may be confusing to others, but it is correct.  I left it out on accident when I originally committed the file.

mirrormanager2/lib/notifications.py
```diff
@@ -39,11 +39,13 @@ def fedmsg_publish(*args, **kwargs):  # pragma: no cover
...
-        fedmsg.publish(*args, **kwargs)
-    except Exception, err:
-        warnings.warn(str(err))
+    #try:
+    #    import fedmsg
+    #    fedmsg.publish(*args, **kwargs)
+    #except Exception, err:
+    #    warnings.warn(str(err))
+    import fedmsg
+    fedmsg.publish(*args, **kwargs)
```
ralphbean: I think the changes here in ``notifications.py`` should be left out.  The original idea was to make fedmsg optional only for people who really wanted to use it.  Could you revert just this file, @Devyani-Divs?

ralphbean:
This looks like an excellent start.  :)

mirrormanager2/app.py
```diff
@@ -373,14 +375,22 @@ def host_new(site_id):
 
         try:
             SESSION.flush()
+            status = 'Host added'
             flask.flash('Host added')
         except SQLAlchemyError as err:
             SESSION.rollback()
+            status = 'Could not create the new host'
```
pypingou: Do we send a message when we fail to do an action?

mirrormanager2/app.py
```diff
@@ -373,14 +375,22 @@ def host_new(site_id):
 
         try:
             SESSION.flush()
+            status = 'Host added'
```
pypingou: I fear this might be a little confusing, the topic is ``mirror.added`` and the status says ``Host added``, maybe we can just change the topic to represent the action ``host.added``, what do you think?

mirrormanager2/app.py
ralphbean: As it is written, it looks like yes.  But I guess I'd advise against it in this case.

mirrormanager2/app.py
ralphbean: Agreed.  ``host.added`` is more accurate.

mirrormanager2/lib/notifications.py
Devyani-Divs: @ralphbean , yup, doing it right away. :)

mirrormanager2/app.py
Devyani-Divs: @ralphbean @pypingou removing the status messages to be sent when it fails to do an action.

fedmsg.d/example-config.py
Devyani-Divs: @ralphbean , ok. :)

mirrormanager2/app.py
Devyani-Divs: @ralphbean @pypingou yes, updated to "host.added". Also, updated all the edits as discussed. Please review them. Thankyou :)

pypingou: Now that the topic has been adjusted, do we need the status? :)

mirrormanager2/app.py
```diff
@@ -380,7 +383,13 @@ def host_new(site_id):
             APP.logger.debug('Could not create the new host')
             APP.logger.exception(err)
 
+        message = dict(
+            status=status,
```
pypingou: If an error occurred when saving the host to the database, should we send the message on fedmsg?

ralphbean: Hm, after thinking about it a bit it seems like the answer here is probably to not send the message.

@Devyani-Divs, could you move the publish line up inside the ``try/except`` block above just after the ``flask.flash`` line?  That way, if the ``SESSION.flush()`` line produces some sort of error, then we won't publish a message that says "a host was added" when the host was not actually added due to some error.

Devyani-Divs: @pypingou @ralphbean , am so sorry. dont know how i missed that. Will be more careful next time.

ralphbean:
Looks good to me now.  :+1:.  @pypingou, do you want to have a last look also?

pypingou:
It looks good to me, the only question I have is if we want more content in the message itself, but this is something we can improve later on (until the code is released).

ralphbean:
Yeah.  Adding to messages after the fact is easy.  (removing or restructuring causes more pain)

ralphbean:
Thanks for all your work on this @Devyani-Divs!

Devyani-Divs:
@ralphbean @pypingou Thankyou for helping me with my naive doubts. :)

PR #17, opened 2015-03-17
Fix host selection logic in the crawler.

adrianreber:
If any of the active fields is not set the host is disabled;
Not the other way around.


adrianreber:
The current crawler was only crawling deactivated hosts instead of only active hosts.

ralphbean:
Looks good to me.  @pypingou?

pypingou:
:+1: as well

PR #18, opened 2015-03-17
Print out the rsync command (including parameters)

adrianreber:
For better problem analysis it is helpful to know
which rsync command is actually used by MirrorManager.

### DIFF
--- a/utility/mm2_update-master-directory-list
+++ b/utility/mm2_update-master-directory-list
@@ -623,7 +623,7 @@ def parse_rsync_listing(session, config, category, f):
 
 def sync_directories_using_rsync(session, config, rsyncpath, category):
     try:
-        result, output = run_rsync(rsyncpath, category.extra_rsync_options)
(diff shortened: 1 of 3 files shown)


adrianreber:
This is something which I was always looking for. Which rsync command is actually used by MirrorManager for the crawling. I am not sure if this is this is the correct python way of doing things but it seems to work for my tests.

utility/mm2_update-master-directory-list
```diff
@@ -622,8 +622,9 @@ def parse_rsync_listing(session, config, category, f):
 
 
 def sync_directories_using_rsync(session, config, rsyncpath, category):
+    global logger
```
ralphbean: No need for the ``global logger`` statement here.

adrianreber:
I updated this PR and removed the ```global logger``` line.

pypingou:
Cool, thanks!

PR #20, opened 2015-03-18
Add and install a config file for systemd so that it creates /var/run/mirrormanger upon reboot

pypingou:


pypingou:
Fixes https://fedorahosted.org/mirrormanager/ticket/56

nirik:
Looks good to me. +1

pypingou:
Thanks @nirik

PR #21, opened 2015-03-19
mark_not_up2date expects a host not host.id

adrianreber:


adrianreber:
This is a minimal fix. When running into timeouts (120 minutes) the crawler was exiting with:

```
Traceback (most recent call last):
  File "/home/fedora/adrian/mm2_crawler", line 1024, in <module>
    sys.exit(main())
  File "/home/fedora/adrian/mm2_crawler", line 165, in main
    doit(options, config)
  File "/home/fedora/adrian/mm2_crawler", line 91, in doit
  File "/usr/lib64/python2.7/multiprocessing/pool.py", line 250, in map
    return self.map_async(func, iterable, chunksize).get()
  File "/usr/lib64/python2.7/multiprocessing/pool.py", line 554, in get
    raise self._value
AttributeError: 'int' object has no attribute 'set_not_up2date'
```
One parameter was wrong: host.id instead of host

pypingou:
Thanks for fixing this!

PR #22, opened 2015-03-19
Enable per crawler thread logging

adrianreber:
MM1 used to log for each crawler the output under <hostid>.log.
This adds the per host logging to the new thread based design.
These log files are referenced on each host page in the web
interface.


utility/mirrormanager2.spec
```diff
@@ -219,6 +227,8 @@ cp -r utility/zebra-dump-parser $RPM_BUILD_ROOT/%{_datadir}/mirrormanager2/
 
 
 %files crawler
+%config(noreplace) %{_sysconfdir}/logrotate.d/mm2_crawler
+%{_localstatedir}/log/mirrormanager
```
pypingou: Nit picking but could we add a trailing '/' to indicate it's a folder?

pypingou:
One remark, otherwise, very cool :+1:

adrianreber:
Branch updated with a trailing slash in the spec file.

pypingou:
Awesome :)

PR #23, opened 2015-03-19
Do not run cronjobs as root

adrianreber:
To be able to not run the cronjobs as root but as the
user 'mirrormanager' this user is now created for the
mirrorlist, backend and crawler sub-package. The required
directories are now also owned by mirrormanager:mirrormanager.


adrianreber:
First try to run the cronjobs not any longer as root.

pypingou:
Looks good to me :+1:

PR #24, opened 2015-03-20
Update the last_crawled and last_crawled_duration correctly.

adrianreber:
The update of the field last_crawled was not working due to the
wrong name (lastCrawled).
The field last_crawl_duration was not updated at all as it used
to be written by the master process doing the forking (before using
threads).
Only the changes to the threaded DB session where commit()ted and
close()d but the ones to the global session. To update crawl time
and duration the global DB session is now also commit()ted and
close()d.


pypingou:
Cool, :+1:

PR #25, opened 2015-03-20
Fix systemd's tempfile.conf for mirrormanager2

pypingou:


nirik:
I'm sorry I missed this when looking at the last change. ;( 

+1 this time for real...

pypingou:
We added the mirrormanager user afterward so I don't think you could have foreseen it ;-)

Thanks for the review

PR #26, opened 2015-03-20
Fix link to the crawler log file

adrianreber:
Simple fix to link to the correct crawler log file.

### DIFF
--- a/mirrormanager2/templates/fedora/host.html
+++ b/mirrormanager2/templates/fedora/host.html
@@ -46,7 +46,7 @@
 <p>
 Last Checked In: {{ host.last_checked_in or '-' }} <br />
 Last Crawled: {% if host.last_crawled %} {{ host.last_crawled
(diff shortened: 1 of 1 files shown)


ralphbean:
Sure.  :+1: here.

Although, there if often a better way to use ``url_for(...)`` by passing some arguments in.  That may not work in this case.  Do you know what I mean @adrianreber?

adrianreber:
@ralphbean I was looking for a way to get the base URL of the application. This seemed like one way to get it, but if there is better way to retrieve the base URL I would be happy to learn about it.

ralphbean:
> if there is better way to retrieve the base URL

No, this is the best way to retrieve the base url.  However, you can use ``url_for`` to retrieve *specific* urls that are provided by the flask application.  I see now that the crawler host logs, however, are not served by the flask application and would instead be served by apache, so my suggestion is not relevant here.

:+1: to merge.

pypingou:
:+1: as well

PR #27, opened 2015-03-21
Close per thread logging correctly.

adrianreber:
If the crawler was running for some time the per thread
file logging would start to log to multiple files. Now,
with the removal and closing of the FileHandler the
logging works as it is expected to.


pypingou:
:+1:

PR #28, opened 2015-03-23
Add more informations to the log output.

adrianreber:
Add information about the number of active threads and
how many hosts still need to be crawled to the crawler
logging output.


adrianreber:
The logging output now looks like:

INFO:crawler:Hosts(428/435):Threads(15/75):hostid:hostname:message

ralphbean:
Looks great!

pypingou:
Pretty cool!

PR #29, opened 2015-03-23
Start crawling the hosts which require the most time.

adrianreber:
MM1 used to sort the hosts to be crawled by the duration
of the last crawl (descending). Add this functionality to MM2.


adrianreber:
I am not sure that this is correct. I can see that the order_by is applied but the result from the query is not ordered. If I just print the result of the query the result seems to be in no order at all. Any ideas what I am doing wrong? Is there another approach to get the result of a ordered query?

adrianreber:
I found it. There is another order by statement applied before mine.

PR #31, opened 2015-03-23
Let the database to the work

adrianreber:
As the functionality is already there only select the hosts
from the database which will actually be used later. By not
selecting anyway unwanted hosts the memory usage should be
also slightly decreased.


pypingou:
Good idea, :+1:

PR #32, opened 2015-03-24
This commit combines a few simple changes.

adrianreber:
* Initialize the stats dict correctly.
* Store the time of the last crawl also if the crawl fails.
* Commit the database changes also if an error occured.

### DIFF
--- a/utility/mm2_crawler
+++ b/utility/mm2_crawler
@@ -516,9 +516,8 @@ def report_stats(stats):
 
 def sync_hcds(session, host, host_category_dirs):
     stats = dict(up2date = 0, not_up2date = 0, unchanged = 0,
-                 unknown = 0, newdir = 0, deleted_on_master = 0)
+                 unknown = 0, newdir = 0, deleted_on_master = 0, duration = 0)
     current_hcds = {}
-    host.last_crawled = datetime.datetime.utcnow()
     host.last_crawl_duration = time.time() -  threadlocal.starttime
     stats['duration'] = host.last_crawl_duration
     keys = host_category_dirs.keys()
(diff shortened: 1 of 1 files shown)


adrianreber:
This PR includes some simple one-line changes packed together.

utility/mm2_crawler
```diff
@@ -1067,8 +1063,11 @@ def worker(options, config, host):
         rc = 3
 
     logger.info("Ending crawl of %r with status %r" % (host, rc))
+    host.last_crawled = datetime.datetime.utcnow()
     logger.removeHandler(fh)
     fh.close()
+    session.commit()
```
ralphbean: Are you sure we want to commit even if an error occurred?  Typically, you don't want to do this as it might mean the database records are in some kind of inconsistent or unexpected (exceptional) state.

adrianreber:
Hmm... The reasoning behind this change was that the changes were only committed after a successful per_host() run. The changes were not committed for the case TimeoutException:.

As the last_time_crawled was not updated correctly in the mark_not_up2date() function I thought I move the commit() and last_time_crawled at the end so that it is done at one central point. Your are right, that it should probably not be committed if something failed.

So I will change this to only commit and write last_time_crawled in try: and TimeoutException:
The close can be left at the place I put it I think. I will update this PR with a new version.

adrianreber:
I have updated my branch and I am not only committing the changes if the crawl was successful or if it timed out. Not if it failed for another reason. I also moved the last_time_crawled just before the session.commit() in both cases session.close() is still after the exception handling.

ralphbean:
Looks good to me now.  :+1:!

PR #34, opened 2015-04-02
Ask to build with systemd-devel to simplify

pypingou:
This will bring the definition of the RPM macros used in %files:
%{_tmpfilesdir}/ and %{_unitdir}/


ralphbean:
:+1:

PR #36, opened 2015-04-09
Additional columns

adrianreber:
If MM2 should go productive soon I would like to add additional fields to the database so that additional crawler features do not need changes to the database.

### DIFF
--- a/alembic/versions/24681dabe5fa_add_host_columns.py
+++ b/alembic/versions/24681dabe5fa_add_host_columns.py
@@ -1,0 +1,31 @@
+"""add_host_columns
+
+Revision ID: 24681dabe5fa
+Revises: None
+Create Date: 2015-04-09 11:06:47.978441
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '24681dabe5fa'
+down_revision = None
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column('host', sa.Column('crawl_failures', sa.Integer))
+    op.add_column('host', sa.Column('disable_reason', sa.Text()))
+    op.add_column('host', sa.Column('push_ssh_private_key', sa.Text()))
+    op.add_column('host', sa.Column('push_ssh_host', sa.Text()))
+    op.add_column('host', sa.Column('push_ssh_command', sa.Text()))
+    op.add_column('host', sa.Column('last_crawls', sa.PickleType()))
+
+def downgrade():
+     op.drop_column('host', 'crawl_failures')
+     op.drop_column('host', 'disable_reason')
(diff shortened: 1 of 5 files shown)


mirrormanager2/lib/model.py
```diff
@@ -193,6 +193,24 @@ class Host(BASE):
...
+    # This can be used to auto disable a host if the crawler fails
+    # multiple times in a row.
+    crawl_failures = sa.Column(sa.Integer, default=0, nullable=False)
+    # Add a text field to specify why the mirror was disabled.
+    # This can either be filled by the crawler for auto disable due
+    # to crawl failures, or by an admin (e.g., ticket number)
+    disable_reason = sa.Column(sa.Text(), nullable=True)
+    # If SSH based push mirroring will ever be implemented
+    # this field should contain the public key of the destination host
+    push_ssh_public_key = sa.Column(sa.Text(), nullable=True)
```
pypingou: Don't we need the private key if we want to push there?

pypingou:
There is a ``..pyo`` file in there that could be dropped I think.

adrianreber:
Yes, private key, of course. I will update this PR.

There is also a .pyc file. Can that also be dropped?

pypingou:
I think so for the ``.pyc`` yes

adrianreber:
PR udpated

alembic/versions/24681dabe5fa_add_host_columns.py
```diff
@@ -0,0 +1,31 @@
...
+down_revision = None
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column('host', sa.Column('crawl_failures', sa.Integer))
+    op.add_column('host', sa.Column('disable_reason', sa.Text()))
+    op.add_column('host', sa.Column('push_ssh_public_key', sa.Text()))
```
pypingou: That one still says public_key

alembic/versions/24681dabe5fa_add_host_columns.py
```diff
@@ -0,0 +1,31 @@
...
+    op.add_column('host', sa.Column('push_ssh_host', sa.Text()))
+    op.add_column('host', sa.Column('push_ssh_command', sa.Text()))
+    op.add_column('host', sa.Column('last_crawls', sa.PickleType()))
+
+def downgrade():
+     op.drop_column('host', 'crawl_failures')
+     op.drop_column('host', 'disable_reason')
+     op.drop_column('host', 'push_ssh_public_key')
```
pypingou: And here as well :)

adrianreber:
Thanks for spotting these. Updated.

pypingou:
:+1: for me

PR #38, opened 2015-04-12
Fix crawler timeout

adrianreber:
If the given timeout for the crawling of a host has been reached
has not been checked at enough places. Now the crawler actually
aborts if a host has not been crawled in the given time.
The timeout_check() function no longer gets the complete options
data structure but only the timeout and a few additional comments
have been added to the code.

Previously the crawler marked a host as not up to date if a single
category on that host has time out. This functionality is still the
same but now explicitly mentioned in the comments.


pypingou:
Looks good to me :+1:

PR #41, opened 2015-04-14
Add a Location tag for static

puiterwijk:
For every alias, you need to have either a Directory or a Location
that matches it.
Source: https://wiki.apache.org/httpd/ClientDeniedByServerConfiguration


pypingou:
wfm :+1:

pypingou:
Well, actually, do you know if there is anything at this location? Looking at the spec file I even wonder if the location exists and I do not know what static files the mirrorlist could even serve.

Do you know?

puiterwijk:
Yes: there's a /static/css/fedora.css, /static/css/main.css and /static/css/style.css that are actually used in the mirrorlist homepage: http://mirrors.fedoraproject.org/publiclist/

PR #43, opened 2015-04-14
Decrease time required for set_not_up2date()

adrianreber:
Now that there are more checks if a timeout has occurred it was
observed that the function set_not_up2date() takes enormous amounts
of time to finish. The function is looping over each directory in
a category and marking it as not up to date. Unfortunately there
was a session.commit() after each directory update which required
huge amounts of time. Now all directories are marked as not up to
date and then a single session.commit() is executed.

This is still a bit strange as all other DB accesses and changes
are only committed at the end of a crawler thread and not somewhere
in a function which is outside of the crawler.


pypingou:
Sounds good :+1:

PR #44, opened 2015-04-15
Add support to auto disable mirrors

adrianreber:
Mirrors which have failed to be successfully crawled for a certain
number of crawls are now automatically disabled (user_active).
The reason for the automatic disabling is visible in the field
disable_reason.


pypingou:
:+1: for me

PR #45, opened 2015-04-15
Auto disable hosts which have a URL configured but which does not exist

adrianreber:
There are a few hosts in the DB which do not have the configured URL
anymore. Auto disable the hosts if the broken URL persists for a
certain number of crawls. Before any other access to the host a
connection is made if either via http or ftp the host can be reached.

### DIFF
--- a/utility/mm2_crawler
+++ b/utility/mm2_crawler
@@ -890,6 +890,43 @@ def select_host_categories_to_scan(session, options, host):
         result = list(host.categories)
     return result
 
+def check_for_base_dir(hoststate, urls):
+    """Check if at least one of the given URL exists on the remote host.
+    This is used to detect mirrors which have completely dropped our content.
+    This is only looking at http and ftp URLs as those URLs are actually
+    relevant for normal access. If both tests fail the mirror will be marked
+    as failed during crawl.
+    """
+    filedata = { 'size' : 0 }
+    exists = False
+    for u in urls:
+        if u.startswith('http:'):
+            try:
+                exists = check_head(hoststate, u, filedata, False, True)
+            except:
+                exists = False
+            if not exists:
+                logger.warning("Base URL %s does not exist." % u)
+                continue
+            # The base http URL seems to work. Good!
+            return True
+        if u.startswith('ftp:'):
+            try:
+                exists = get_ftp_dir(hoststate, u, True)
+            except TryLater:
+                # Some temporary difficulties on the mirror
+                exists = False
+            # exists is an empty list if that directory does not exist
+            if not exists:
+                logger.warning("Base URL %s does not exist." % u)
+                continue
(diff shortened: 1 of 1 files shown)


utility/mm2_crawler
```diff
@@ -935,6 +966,17 @@ def per_host(session, host, options, config):
             continue
         categoryPrefixLen = len(category.topdir.name)+1
 
+        all_categories_have_failed = False
```
pypingou: This one is already instantiated above. Is that intended?

utility/mm2_crawler
```diff
@@ -1026,12 +1068,28 @@ def per_host(session, host, options, config):
                 % (host.name, host.id))
     hoststate.close()
 
+    if all_categories_have_failed:
+	# Let's say that '5' is the signal for the calling function
```
pypingou: Indentation error here

utility/mm2_crawler
```diff
@@ -1026,12 +1068,28 @@ def per_host(session, host, options, config):
...
             sync_hcds(session, host, host_category_dirs)
     return rc
 
+def count_crawl_failures(host, config):
+   try:
+       host.crawl_failures += 1
+   except TypeError:
+       host.crawl_failures = 1
 
+   auto_disable = config.get('CRAWLER_AUTO_DISABLE', 5)
```
pypingou: Isn't the default 4 in ``default_config.py``?

utility/mm2_crawler
```diff
@@ -1026,12 +1068,28 @@ def per_host(session, host, options, config):
...
+   if host.crawl_failures >= auto_disable:
+       host.disable_reason = "Host has been disabled (user_active) after %d consecutive crawl failures" % auto_disable
```
pypingou: Longer than 80 chars it seems :)

utility/mm2_crawler
```diff
@@ -1026,12 +1068,28 @@ def per_host(session, host, options, config):
...
+       host.user_active = False
```
pypingou: Discussing this with @ralphbean we thought that it would be nice to have a way to notify the mirror admin about this. One way would be to rely on fedmsg and FMN to notify the mailing list. What do you think?

adrianreber:
Thanks for the review. Let me think about a few of your questions.

adrianreber:
I tried to fix all the problems you have found. Concerning the notification of the auto disabled mirror. We have an email address from the mirrors called 'robot' which could be used or we can try the data from FAS. Or fedmsg. All options sound okay to me. Whatever makes most sense.

pypingou:
:+1: for me

I think I like the notification via fedmsg, it would allow us to send them to the list and avoid writing yet another email sending block of code, but up to you :)

PR #46, opened 2015-04-20
crawl_duration is a host specific property

adrianreber:
Only update crawl_duration at the end of crawl of a host
and not after each category. As it is a host specific property
it only makes sense to update it once.


pypingou:
:+1: for me

PR #47, opened 2015-04-20
Handle lighttpd returing a content length for directories

adrianreber:
lighttpd returns a Content-Length for directories.
apache and nginx do not.
For the basic check in check_for_base_dir() it is only
relevant if the directory exists or not. Therefore
passing None as filedata[]. This needs to be handled.


pypingou:
wfm :+1:

PR #48, opened 2015-04-21
Scan the directories which are supposed to be on each mirror.

adrianreber:
The crawler was only looking at the files which a mirror actually
had and not at the files it is supposed to have. This lead to the
situation that a newly added host or category was never crawled
successfully. Now the crawler checks the list of files which are
supposed to be on the host and compares it with the actual content
of the crawled mirror.


pypingou:
Looks simple, almost too simple for the description in the commit message, but code-wise :+1: for me :)

PR #49, opened 2015-04-21
Use yesterday's files to avoid any time zones troubles

adrianreber:


adrianreber:
This should silence the emails from cron.

pypingou:
what about using ``date -u`` instead?

adrianreber:
Ah, good idea. Let me respin.

adrianreber:
No, actually, the problem is not our system which already is on UTC. The problem is the remote system with the Internet2 tables which does not create the new tables because they seem to be on their local time which is multiple hours behind UTC. Looking at the URL now it has the right tables for "today" but the tables were not there a few hours ago. So my current patch still seems to be right fix.

pypingou:
ok, mind adding this as a comment in the file before merging then?

Otherwise :+1: for me

PR #50, opened 2015-04-22
Umdl log and category fixes

adrianreber:
These two commits try to fix logging to umdl.log and where that file should be and also introduces the possibility to only scan a single category. This opens up the possibility to start a category specific scan upon fedmsg notifications and not run a scan over the whole tree every two hours.

### DIFF
--- a/mirrormanager2/default_config.py
+++ b/mirrormanager2/default_config.py
@@ -78,6 +78,13 @@ MM_COOKIE_REQUIRES_HTTPS = True
 # Default: ``.MirrorManager``.
 MM_COOKIE_NAME = 'MirrorManager'
 
+# If this variable is set (and the directory exists) the crawler
+# will create per host log files in MM_LOG_DIR/crawler/<hostid>.log
+# which can the be used in the web interface by the mirror admins.
+# Other parts besides the crawler are also using this variable to
+# decide where to store log files.
+MM_LOG_DIR = '/var/log/mirrormanager'
+
 # If not specified the application will rely on the root_url when sending
 # emails, otherwise it will use this URL
 # Default: ``None``.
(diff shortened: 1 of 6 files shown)


mirrormanager2/default_config.py
```diff
@@ -78,6 +78,13 @@
 # Default: ``.MirrorManager``.
 MM_COOKIE_NAME = 'MirrorManager'
 
+# If this variable is set (and the directory exists) the crawler
+# will create per host log files in MM_LOG_DIR/crawler/<hostid>.log
+# which can the be used in the web interface by the mirror admins.
```
pypingou: Hm, not sure apache will be allowed by SELinux to go there

utility/mm2_update-master-directory-list
```diff
@@ -758,19 +800,29 @@ def main():
...
-    for i in config.get('umdl_master_directories'):
+    check_categories = []
+    if options.categories is None:
+        check_categories = config.get('umdl_master_directories')
+    else:
+        for i in config.get('umdl_master_directories'):
+            if i['category'] == options.categories:
+                check_categories.append(i)
+
+    print check_categories
```
pypingou: Forgot to remove this line or is it meant to stay?

mirrormanager2/default_config.py
adrianreber: No, I had to change the SELinux context of that directory on the staging crawler.

utility/mm2_update-master-directory-list
adrianreber: Will remove that line. You are right, that was only for debugging.

mirrormanager2/default_config.py
pypingou: Might be something worth adding to the ansible playbook :)

pypingou:
:+1: once the print statement is removed

PR #54, opened 2015-04-26
[RFC] Trying explicit garbage collection

adrianreber:
The crawler requires a lot of memory. With the explicit garbage
collection at the end of each crawler thread the crawler has not
been OOM killed for the last 4 runs. If this actually helps is
not very clear but it seems to slightly decrease the OOM killed
frequency. Unfortunately it seems difficult to tell which part
of the crawler and which thread is using the memory and for what.

### DIFF
--- a/utility/mm2_crawler
+++ b/utility/mm2_crawler
@@ -16,6 +16,7 @@ import threading
 import time
 import urllib2
 import urlparse
+import gc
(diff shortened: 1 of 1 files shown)


pypingou:
Clean and easy to revert :+1:

PR #55, opened 2015-04-27
Crawler documentation

adrianreber:


doc/crawler.rst
```diff
@@ -0,0 +1,91 @@
...
+
+The crawler retrieves the list of active mirrors from the database and
+spawns a thread for each host. The crawling of a host consists of a loop
+over all categories which have been configured for this host and one category
+is crawled after another.
+
+The crawler has the possibility to crawl a category via HTTP, FTP or RSYNC.
+For each category the protocol to crawl is selected from the available
+category protocols. RSYNC has the highest priority and is followed by HTTP
+and the protocol used last (if no other protocol is available) is FTP.
```
pypingou: Stupid question, but do we know if there is a reason for this order?

doc/crawler.rst
```diff
@@ -0,0 +1,91 @@
...
+and the timestamp of the crawl is recorded in the host. After the crawl of
+each category the files which are up to date in the category are stored
+in the database. Crawl failures (timeouts, network problems) usually result
+in a complete 'marked-as-not-up-to-date' of that mirror. Multiple consecutive
+crawl failures (default 4) will disable the host completely (host.user_active).
+
+The crawler requires enormous amounts of memory and for 40 threads crawling
+mirrors in parallel at least 32GB of memory are required. At the end of
+each crawl thread the garbage collector is explicitly called in the hope
+that some unused memory is freed again.
```
pypingou: We may want to specify the amount of hosts/files crawled as I guess another system (CentOS RPMFusion...) will have different numbers/needs there

doc/crawler.rst
```diff
@@ -0,0 +1,91 @@
...
+
+Crawling protocol
+-----------------
+
+As previously mentioned the crawler uses either RSYNC, HTTP or FTP to crawl
+a connected mirror. As also already mentioned the crawler spawns a thread
+for each host it has to crawl and then starts crawling one category configured
+after another. For each category the protocol decision made anew.
+
+It the category supports RSYNC the whole category is scanned using RSYNC with
```
pypingou: if?

doc/crawler.rst
```diff
@@ -0,0 +1,91 @@
...
+each crawl thread the garbage collector is explicitly called in the hope
+that some unused memory is freed again.
+
+Crawling protocol
+-----------------
+
+As previously mentioned the crawler uses either RSYNC, HTTP or FTP to crawl
+a connected mirror. As also already mentioned the crawler spawns a thread
+for each host it has to crawl and then starts crawling one category configured
+after another. For each category the protocol decision made anew.
```
pypingou: is made anew?

doc/crawler.rst
```diff
@@ -0,0 +1,91 @@
...
+
+It the category supports RSYNC the whole category is scanned using RSYNC with
+a single network connection. If it was able to find all files the category
+is marked as up to date and the next category follows. If no RSYNC URL is
+available the crawler uses FTP or HTTP. FTP requires one network connection
+per directory and using HTTP each file is crawled separately. Depending on
+the configuration of the remote host this can mean one network connection
+per file or, if the remote host supports HTTP keep-alive, one connection
+for multiple files or the whole category. This depends again on the
+configuration of the remote host.
```
pypingou: and is detected automatically?

doc/crawler.rst
```diff
@@ -0,0 +1,91 @@
...
+
+Using HTTP most of the files are only read via HTTP HEAD. Not the actual data
+of the file is downloaded. Only the metadata is downloaded. Only files with
+the name 'repomd.xml' are actually completed downloaded and their SHA256 sum
+is compared to the one in the database.
```
pypingou: don't we also download some files for the atomic tree?

pypingou:
Few questions but this is nice and clear, thanks!

doc/crawler.rst
adrianreber: Good question about the protocol order. This is probably related that most mirrors support HTTP and that FTP is not the best protocol when it comes to firewalls. Looking at the code of per_host() the crawler first tries to scan a whole category using RSYNC and if that fails does a whole directory using FTP and the last option is each single file via HTTP.
But then there is the function which selects the preferred protocol which has the ordering RSYNC -> HTTP -> FTP. I think this function could be from me.  The reason I put FTP at last probably comes from dislike of FTP as a mirror admin. I would rather get rid of it. But maybe this needs to be re-evaluated.

doc/crawler.rst
adrianreber: Looking at the code of the crawler, atomic is not mentioned at all. This is probably one of the reasons crawling takes so long. The atomic tree needs to be handled differently.

PR #56, opened 2015-04-28
Also add a /var/run directory for the backend

adrianreber:
To be able to use fedmsg based umdl runs there needs to be a place to store the last time umdl was run. Using /var/run/mirrormanager in the backend package seems like the right thing to do.


pypingou:
wfm :+1:

PR #57, opened 2015-04-28
Add fedmenu

pypingou:


oddshocks:
:+1:

ralphbean:
mirrormanager2 has a concept of users, right?  (they are mirror admins)

Is there a page to view them and their details?  if so, we should consider using the ``'user': '$USERNAME',`` feature of fedmenu.

pypingou:
On Wed, Apr 29, 2015 at 04:16:40PM -0700, Ralph Bean wrote:
>    mirrormanager2 has a concept of users, right? (they are mirror admins)

yes in the sense that users log in and can be added to a mirror

>    Is there a page to view them and their details? if so, we should consider
>    using the 'user': '$USERNAME', feature of fedmenu.

There are no such page actually :)

ralphbean:
Then :+1:!  :dancers:

PR #58, opened 2015-04-30
Add new utility script to be used to create the fedora-install-X repositories

pypingou:


ralphbean:
This looks okay to me, but it's one of those things that will just need to be tested before we can say anything certain about it.

pypingou:
Seems to run fine like this in stg, but not sure how to check it

adrianreber:
To test it you need to copy the files from /var/lib/mirrormanager (after they have re-created by mm2_update-mirrorlist-server) to the mirrorlist server and kill -1 the mirrorlistserver process

pypingou:
hm, I'm not sure I follow you, it's a one off script taking data from the disk and adding it to the DB.

To run it, on mm-backend01.stg, I have been using:
``
mm2_create_install_repo --version=21 --category='Fedora Linux' /srv/pub/fedora/linux/releases/21/
``
which seemed logical and seems to have added things in the DB: https://admin.stg.fedoraproject.org/mirrormanager2/admin/repositoryview/?page=57 but apparently there is still something missing

pypingou:
Ok I think I got it: https://admin.stg.fedoraproject.org/mirrormanager2/admin/repositoryview/?sort=5&page=15

pypingou:
I ran it with 
``
cd /srv/ && mm2_create_install_repo --version=21 --category='Fedora Linux' pub/fedora/linux/releases/21/ 
``

mdomsch:
Why  is that not done in umdl like the other repos?

adrianreber:
@mdomsch: I guess this is 'historically' (as always). It could be part of umdl but it needs different paths starting with Fedora 21:

fedora/linux/releases/21/Server/x86_64/os/
fedora/linux/releases/20/Fedora/x86_64/os/

This is probably worth a future umdl feature. I will open a ticket to integrate it into umdl.

pypingou:
So let's merge this for now so that we can cut a new release :)

PR #62, opened 2015-05-05
Fix version handling

adrianreber:
The wrong version was retrieved from the database and thus the new
repositories were pointing to some bogus version. Now first the
category is retrieved from the database and then with the category
and the product name the correct version is retrieved from the
database.

Additionally the parameter handling was optimized and wrong comments
were removed.

### DIFF
--- a/utility/mm2_create_install_repo
+++ b/utility/mm2_create_install_repo
@@ -2,20 +2,6 @@
 
 """
 This script creates the fedora-install repo for Fedora .
-
-Basically, it checks the content of the nfs mount, if the directory
-contains a lot of files, it takes the 3 most recents ones, stores them in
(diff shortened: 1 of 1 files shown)


utility/mm2_create_install_repo
```diff
@@ -143,6 +123,14 @@ def main():
         "--debug", dest="debug", action="store_true", default=False)
 
     (options, args) = parser.parse_args()
+
+    if len(args) != 1:
+        parser.error("Must specify the path to the base of the install repo\n"
+                     "for example: mm2_create_install_repo --version=21 "
+                     "--category="Fedora Linux" pub/fedora/linux/releases/21/")
```
pypingou: The quotes around Fedora Linux here will be problematic

pypingou:
:+1:

PR #63, opened 2015-05-05
Fix pickle generation when several repositories point to the same directory

pypingou:


adrianreber:
Tested on the MM2 production systems and now the correct fedora-install-21 repositories are generated. This fixes our problem. :+1:

There is still the question, however, why are there repositories without a prefix and the same directory_id.

PR #64, opened 2015-05-07
Allow deleting sites and hosts and cascade the change

pypingou:


adrianreber:
Looks good, but I cannot say much about the sqlalchemy cascading requests. If your tests have been successful I would say :+1:

PR #65, opened 2015-05-07
Do not only create /var/lock/mirrormanager on installation.

adrianreber:
So that /var/lock/mirrormanager is created on every reboot it also needs
to be included in tmpfiles.d.


pypingou:
:+1:

ralphbean:
:+1:

PR #66, opened 2015-05-08
Include the header in the fedora template directly

adrianreber:
Do not load the background image from koji.fedoraproject.org anymore
over http. Include it directly in the static/ directory.


pypingou:
:+1:

PR #67, opened 2015-05-08
Support always update hosts which are unreachable.

adrianreber:
The crawler was not able to reach dl.fedoraproject.org due to
firewalls and has therefore disabled the master mirror. Now all
categories which are marked as always up to date count as a
successfully scanned category and are not auto-disabled after a
short time.


pypingou:
fun bug :)

:+1:

PR #68, opened 2015-05-10
Adjust the spec file to the systemd packaging guidelines for Fedora

pypingou:
reference: https://fedoraproject.org/wiki/Packaging:ScriptletSnippets#Systemd


ralphbean:
:+1:

adrianreber:
:+1:

btw. I see BuildRequires: systemd-devel in the specfile. The wiki page you mentioned says to BR systemd.

pypingou:
Regarding systemd-devel I seem to remember we needed it for a macro (but I don't remember which one, I'll test and adjust if necessary)

PR #69, opened 2015-05-10
Crawler

adrianreber:
This PR includes a few improvements to the crawler. Especially a canary and repodata crawl mode. New fedmsg topics and better fedmsg information about the crawled hosts.

### DIFF
--- a/utility/mm2_crawler
+++ b/utility/mm2_crawler
@@ -31,8 +31,7 @@ current_host = 0
 all_hosts = 0
 threads = 0
 threads_active = 0
-canary_hosts = 0
-canary_hosts_failed = 0
+hosts_failed = 0
 
 
 # This is a "thread local" object that allows us to store the start time of
@@ -141,8 +140,13 @@ def doit(options, config):
     notify(options, 'complete', dict(results=results))
 
     if options.canary:
-        logger.info("%d of %d hosts disabled in canary mode" %
-                    (canary_hosts_failed, canary_hosts))
+        mode = ' in canary mode'
+    elif options.repodata:
+        mode = ' in repodata mode'
+    else:
+        mode = ''
+
+    logger.info("%d of %d hosts failed%s" % (hosts_failed, current_host, mode))
     return results
 
 
@@ -158,6 +162,7 @@ def setup_logging(debug):
 
 
 def main():
+
     starttime = time.time()
     parser = argparse.ArgumentParser(usage=sys.argv[0] + " [options]")
     parser.add_argument(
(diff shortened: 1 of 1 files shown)


utility/mm2_crawler
```diff
@@ -123,9 +137,24 @@ def doit(options, config):
     return_codes = threadpool.map(fn, hosts)
 
     # Put a bow on the results for fedmsg
-    results = [dict(rc=rc, host=host) for rc, host in zip(return_codes, hosts)]
-    notify(options, 'complete', dict(results=results))
+    results = [dict(rc=rc, host=host, id=id)
+               for rc, host, id in zip(return_codes, host_names, hosts)]
+
+    if options.canary:
+        topic = 'complete_canary'
```
pypingou: I'd use ``complete.canary``

utility/mm2_crawler
```diff
@@ -1003,22 +1042,29 @@ def per_host(session, host, options, config):
...
+            try:
+                has_all_files = try_per_category(
+                    session, trydirs, categoryUrl, host_category_dirs, hc,
+                    host, categoryPrefixLen, options.timeout_minutes, config)
+            except TimeoutException:
+                # If the crawl of only one category fails, the host
+                # is completely marked as not being up to date.
+                raise
+
+            if type(has_all_files) == type(True):
```
pypingou: hm this would return True if has_all_files is False:
```
>>> type(False) == type(True)
True
```

Maybe we can use ``if has_all_files and has_all_files is True`` ?

adrianreber: Interesting. This has always been there. The block has just been indented. The function returning has_all_files returns None, True, False. Looking at it I think it could be changed to only return True or False. I will add an additional commit to this PR to fix this.

adrianreber:
Fixed fedmsg topic and boolean compare.

ralphbean:
:-1: to adding new fedmsg topics for canary mode.

Can you just re-use the existing topics but add a ``canary=True`` or ``canary=False`` to the ``msg`` payload in the ``notify(...)`` function?  FWIW, all kinds of information would be welcome in that payload.. like the values of *all* the options!  You can convert it from an ``argparse.Namespace`` object into a dict like this:

```python
msg['options'] = dict(options._get_kwargs())
```

adrianreber:
Updated the fedmsg part to submit all options without adding additional topics.

ralphbean:
:+1:, thanks Adrian :)

adrianreber:
It looks like I can merge this? Any objections?

ralphbean:
@adrianreber go for it.  :)

Issue #75, opened 2015-05-15
Add possibility to filter mirrorlist by product


adrianreber:
MM1 had the ability to filter the mirrors by product. It was possible to show all EPEL mirrors regardless of version and/or architecture. MM2 has this possibility not. This breaks some external links.

http://mirrors.fedoraproject.org/publiclist/EPEL/

http://mirrors.fedoraproject.org/publiclist/Fedora/

pypingou:
So this is now fixed in git

PR #76, opened 2015-05-15
Offer possibility to sort by product

adrianreber:
The MM1 publiclist had the possibility to not just sort by
product/version and product/version/arch but also only by product. Now
it is possible to only show EPEL or Fedora mirrors. Necessary for links
which point to only a certain product and are now no longer working.

This solves #75


pypingou:
Nice and easy, :+1:

PR #78, opened 2015-05-18
Couple of UI fixes

pypingou:
Fix the label of a submit button from ``Create`` to ``Update`` to reflect what the button does

Only show the ``admin active`` fields to admins


ralphbean:
Looks good, :+1:

adrianreber:
I like the MM1 approach for two reasons:

1. The mirror admins can see that the options exists and it also says that they have to contact one of the MirrorManager admins to get the Peer ASNs listed.
2. The mirror admins **cannot** add Peer ASNs by themselves, which can be dangerous if they start to add the wrong peer ASNs and redirected Fedora clients to mirrors on the other side of the world.

PR #79, opened 2015-05-18
Fix peer ASNs

pypingou:
Only the admins can add/remove them and do not check authorization twice when not needed


pypingou:
@adrianreber this should bring back the behavior you mentionned

adrianreber:
:+1:

PR #80, opened 2015-05-19
Create noauthed master for mirror publiclist

puiterwijk:
This makes it possible to make publiclist varnish cached
because that doesn't depend on any cookie status anymore.

### DIFF
--- a/mirrormanager2/templates/fedora/master.html
+++ b/mirrormanager2/templates/fedora/master.html
@@ -1,96 +1,24 @@
-<!DOCTYPE html>
-<html lang='en'>
-<head>
-    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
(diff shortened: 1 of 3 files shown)


adrianreber:
I cannot say if it is correct but I like the idea very much to be able to cache the mirrorlist as the content does not change that often.

pypingou:
Does not compile:
````
mirrormanager2/mirrormanager2/templates/fedora/master.html", line 3, in template
    {% block nav %}
TemplateSyntaxError: unexpected '%'
````

puiterwijk:
The compile error is now fixed :)

pypingou:
Cool, testing it looks good to me! :)

pypingou:
This way, we can cache everything but the front page (and the pages needing auth), nice and easy! :)

puiterwijk:
Thanks, this should improve it a lot!

PR #87, opened 2015-05-26
Add a new utility script to upgrade repo from -alpha or -beta to release

pypingou:


utility/mm2_upgrade-install-repo
```diff
@@ -0,0 +1,144 @@
...
+        return
+
+    for a in mirrormanager2.lib.get_arches(session):
+        if a.name == u'source':
+            continue
+
+        if options.version == 'development':
+            # We do not change development repos
+            continue
+        else:
```
ralphbean: Since the line before this is a `continue`, is there no need for the `else:` block here?  You could perhaps unindent it and maybe get a little more readability.

utility/mm2_upgrade-install-repo
```diff
@@ -0,0 +1,144 @@
...
+            prefix=u'fedora-install-%s' % ver.name
```
ralphbean: spaces around the ``=`` sign here.

utility/mm2_upgrade-install-repo
```diff
@@ -0,0 +1,144 @@
...
+                    " yet, skipping creation of a repository there until "\
+                    "after the next UMDL run." % d
+                continue
+
+            repo = mirrormanager2.lib.get_repo_prefix_arch(
+                session, prefix, a.name)
+            if not repo:
+                print 'No repo found for prefix: %s - arch: %s' % (
+                    prefix, a.name)
+                return
```
ralphbean: Most of the other error checking sections have ``continue`` instead of ``return``.  Should that be the case here?

pypingou: Def yes, before it was returning a list but now it return a single element so it could continue

ralphbean:
:+1:

pypingou:
Thanks for the review @ralphbean

PR #88, opened 2015-05-27
Adjust the logrotate configuration to fix the permission denied error

pypingou:
we've been seeing

### DIFF
--- a/utility/mm2_umdl.logrotate
+++ b/utility/mm2_umdl.logrotate
@@ -4,6 +4,5 @@
     daily
     dateext
     rotate 15
-    postrotate
-    endscript
+    copytruncate
 }


ralphbean:
I don't know much about logrotate configs off hand.  Can you explain what this fixes and why it fixes it?

pypingou:
This was pointed out by @nirik it's an attempt to fix the permission denied error we've seen in the daily emails from crawler01 and crawler02.

Here is the email from @nirik about this issue:
```
   I think the problem is that the /etc/logrotate.d/mm2_crawler file has:
       postrotate
       endscript

   That defines a postrotate script and passes it the log file as the
   first argument. But there's nothing defined there, so it's just trying
   to run the log file as a command and getting permission denied.

   Is there something here we should be doing to tell currently running
   crawlers to close logs and restart? (httpd has a reload here)

   Or we could drop those 2 lines and replace with:
   copytruncate
   which will just copy the log file to the new archived version and
   truncate the existing one so the crawler can keep appending to it if it
   wants.
```

ralphbean:
Sounds good to me.  :)

Issue #83, opened 2015-05-20
Provide a URL to check for basic functionality


adrianreber:
In environments where MM2 is running behind proxies it would be good to provide a URL which makes a simple DB connection and then returns. The MM2 start page (as opposed to the MM1 start page) does a relatively complex database query which can lead to a high memory and CPU consumption just by multiple proxies checking the availability.

There was the proposal to provide a URL like ```/ping``` which does exactly this and which then just returns OK.

pypingou:
I'm thinking maybe more something like ``/api/repositories`` list each repo by name or something like this, something querying the DB but not too hard and still be somewhat useful

PR #91, opened 2015-05-27
Add a new API endpoint listing the repositories

pypingou:
This endpoint can be used to check the status of mirrormanager's UI

Fixes https://github.com/fedora-infra/mirrormanager2/issues/83

### DIFF
--- a/mirrormanager2/api.py
+++ b/mirrormanager2/api.py
@@ -137,15 +137,15 @@ def api_repositories():
 
     repositories = mmlib.get_repositories(SESSION)
     repos = []
-    for repositorie in repositories:
-        if not repositorie:
+    for repository in repositories:
+        if not repository:
             continue
 
         tmp = {
-            'name': repositorie.name,
-            'directory': repositorie.directory.name,
+            'name': repository.name,
+            'directory': repository.directory.name,
         }
-        repos.append({repositorie.prefix: tmp})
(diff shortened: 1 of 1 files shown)


mirrormanager2/api.py
```diff
@@ -95,3 +95,63 @@ def api_mirroradmins():
...
+          }
+        ]
+        "total": 2
+      }
+
+    '''
+
+    repositories = mmlib.get_repositories(SESSION)
+    repos = []
+    for repositorie in repositories:
```
ralphbean: repository instead of repositorie.

mirrormanager2/api.py
```diff
@@ -95,3 +95,63 @@ def api_mirroradmins():
     jsonout = flask.jsonify(output)
     jsonout.status_code = 200
     return jsonout
+
+
+@APP.route('/api/repositories/')
+@APP.route('/api/repositories')
```
oddshocks: Is something like this always required? Or only in some cases? (just curious, never had to do this)

oddshocks:
Looks proper to me, homedog! :+1:

mirrormanager2/api.py
pypingou: It's not required per say, it's just the way to ask flask to support both URL, with and without the trailing slash. After it's a matter of personal preferences if you use it or not :)

PR #93, opened 2015-06-20
First pass at update-single-file-detail

ralphbean:
This is a script we might try and use to speed up recovery of the pkl data.

### DIFF
--- a/utility/mm2_update-single-file-detail
+++ b/utility/mm2_update-single-file-detail
@@ -1,0 +1,183 @@
+#!/usr/bin/env python
+"""
+This script is like the light-weight version of update-master-directory-list.
+Instead of crawling *all* files on the nfs mount of the master mirror content,
+and updating the mm2 database with what mirrors *should* have.. this script
+looks at just a single master mirror file and updates its details in the
+mm2 database.
+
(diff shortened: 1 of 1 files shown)


adrianreber:
Would it make sense to integrate this functionality directly into umdl?

ralphbean:
> Would it make sense to integrate this functionality directly into umdl?

I tried that first.  It seems like you would only need to replace that ``os.walk`` statement with a fake list including only your filename, but:

- What if you pass it a file that is not one of the ones returned by ``short_filelist``.. I'm not sure how that would work and..
- The logic in ``sync_categories_directories`` confused me when I was trying to modify it for this, so I reversed course and split it out into its own script.

If there are ways to deal with those two things, then yes we could include it in umdl.  :)

ralphbean:
@adrianreber, @pypingou any further feedback here.  If this isn't something we want to merge, we can always :-1: and close it to get it out of the queue.

pypingou:
I'm :+1: to merge and we can always kick it out if we end-up merging it into the UMDL later.

adrianreber:
:+1: what @pypingou said

ralphbean:
Thanks guys!

PR #95, opened 2015-06-20
Umdl delete

adrianreber:
I am not sure this 100% correct, but this are the minimal changes to make umdl's delete option work again. Running umdl with --delete in Fedora's MM staging instance reduces the size of a database dump from 800MB to 760MB.


pypingou:
Simple enough, :+1: for me

PR #102, opened 2015-06-24
If the user is not an admin, keep the existing admin_active settings

pypingou:
Boolean in html are processed in such a way that: present=1 and absent=0
so by hiding the admin_active box to non-admins, we are sending an 'absent'
signal and thus if we do not handle this, we end-up with changing
involuntarily the admin_active flag.


adrianreber:
:+1: Tested in staging and works as expected. This is important as every mirror admin who was looking at the MirrorManager entry disabled his mirror.

Issue #97, opened 2015-06-22
Hide always_up_to_date from non-admins


adrianreber:
The field always_up_to_date should only be accessible by mirrormanager admins in the host category.

PR #103, opened 2015-06-24
Only MM2 admins are allowed to change the always_up2date flag of a mirror

pypingou:
Fixes https://github.com/fedora-infra/mirrormanager2/issues/97


adrianreber:
:+1: Non-admins should not set the mirror to always_up_to_date.

PR #105, opened 2015-07-02
Two simple one line fixes.

adrianreber:
umdl sometimes prints out lines like:

%s: directory already has a repository

this should be fixed and move-to-archive also needed a one line fix.


pypingou:
:+1:

PR #106, opened 2015-07-13
Add support to gracefully shutdown the crawler

adrianreber:
Currently the crawler cannot be gracefully interrupted. This adds code to (more or less) gracefully shut down the crawler threads once SIGARLM has been received.

### DIFF
--- a/utility/mm2_crawler
+++ b/utility/mm2_crawler
@@ -1006,7 +1006,14 @@ def per_host(session, host, options, config):
     hoststate = hostState(
         http_debuglevel=http_debuglevel,
         ftp_debuglevel=ftp_debuglevel,
-        timeout_minutes=options.timeout_minutes)
+        # This used to be the same timout as for the whole category.
+        # This does not make much sense as according to the documentation
+        # ...If the optional timeout parameter is given, blocking operations
+        # ...(like connection attempts) will timeout after that many seconds
+        # ...(if it is not given, the global default timeout setting is used)...
+        # Setting this to '1' should limit the connection establishment to
+        # to 1 minute.
+        timeout_minutes=1)
 
     categoryUrl = ''
     host_categories_to_scan = select_host_categories_to_scan(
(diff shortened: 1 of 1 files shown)


utility/mm2_crawler
```diff
@@ -60,7 +67,17 @@ class InjectingFilter(logging.Filter):
         self.thread_id = thread_id
 
     def filter(self, record):
-        return threadlocal.thread_id == self.thread_id
+        try:
+            return threadlocal.thread_id == self.thread_id
```
pypingou: This can raise an exception?

utility/mm2_crawler
```diff
@@ -126,12 +151,45 @@ def doit(options, config):
...
+            shutdown = False
+        if shutdown_timer:
+            delta = time.time() - shutdown_timer
+            if delta > SHUTDOWN_TIMEOUT * 60:
+                # Time out for graceful shutdown is over.
+                # Let's terminate all threads.
+                logger.warning("About to terminate all threads.")
+                threadpool.terminate()
+                logger.warning("About to terminate all threads. Done.")
+                break
```
pypingou: Aren't we missing a while or so here when we shutdown? Or will ``not result.ready()`` do?

pypingou:
Couple of questions, but I don't think they are blocker so :+1: for me

utility/mm2_crawler
adrianreber: No, but if the shutdown signal comes in too early, before the ```thread_id``` is set, it backtraces here. This is just to make sure early signals are handled properly.

pypingou: ok, thanks

utility/mm2_crawler
adrianreber: Not 100% sure I understand the question. The ```while not result.ready():``` loop will not be left before either all result codes are returned or the loop has been running longer than ```SHUTDOWN_TIMEOUT``` minutes.

pypingou: ok, that's fine then.

Basically, I was wondering if ``result.ready()`` could return ``True`` while we wanted to shutdown and what was going to happen then, but if that's the case, it will just exit fine.

PR #108, opened 2015-07-16
Add support to limit crawling by continent

adrianreber:
With this change a new command-line option is available:

  --continent

It can either be used to specify a continent:

  --continent AF --continent AS

To only crawl mirrors which (according to GeoIP) are either in Africa
Asia. It can also be used to exclude a continent:

  --continent ^EU

Which will scan all mirror which are not in Europe.

The implementation is not optimal which is related to the design of MM.
There is no real location associated with a host. The GeoIP location
can only be done once the categories and the URLs belonging to the
categories are loaded. This means with have to act like we would
actually crawl a host, load the categories and the associated URL.
This implementation now uses the first URL to get (via GeoIP) the
continent this mirror is located. If the mirror is on the desired
continent the mirror is crawled. If, however, the first URL of the
current host is GeoIP-located on another continent the complete host is
skipped. Skipping a mirror due to wrong continent has no negative effect
on the mirror in the database.

### DIFF
--- a/utility/mm2_crawler
+++ b/utility/mm2_crawler
@@ -18,6 +18,7 @@ import time
 import urllib2
 import urlparse
 import gc
+import GeoIP
 
 sys.path.append('..')
 import mirrormanager2.lib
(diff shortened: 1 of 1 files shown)


pypingou:
Looks ok to me :+1:

PR #111, opened 2015-07-22
UMDL2, the rewrite

pypingou:


adrianreber:
Before asking specific questions I would like to ask if you could provide a textual overview about your changes. This would maybe help to understand the changes better and reduce the number of questions I have. The first large commit which moves the code to umdl.py makes reviewing a bit difficult as I do not know if it just moves code or if it also changes code.

pypingou:
So basically, this splits the logic of updating the DB in the umdl.py
The idea is that umdl.py contains the logic that can be used by the different script/method that can be used to update the DB (for example, browser a tree, a filelist or using rsync).
The umdl script itself then just browse the tree and send the info to the umdl.py

So there are some changes to the umdl.py compared to when it was in the umdl script as well (like the method for the repo got changed)

adrianreber:
About the fullfilelist. We once talked about getting more metadata information into this file to further reduce the amount of stat()s we have to do.

adrianreber:
The longer we wait with this PR the more questions I have. Wouldn't it make sense to move umdl.py to mirrormanager2/lib ?

pypingou:
Yes, putting it in mirrormanager2/lib was/is the original idea

adrianreber:
I would prefer if we would first move the code _without_ changes to mirrormanager2/lib/umdl.py (or similar) and once the code has been moved the changes for the new and better umdl2 are applied. I am afraid that if the code is moved and changed at the same time we make it harder on ourselves to keep an overview what has changed an how.

pypingou:
I agree but I pretty much started umdl2 by cutting down everything, I could try to merge back umdl.py into the script and do the split afterward, but no idea how that will work out

pypingou:
Ok I made it so that umdl1 and umdl2 are both present, this way we can release with it and use either but just adjusting the cron job.

What do you think?

adrianreber:
Yes, let's merge it. :+1:  We can start to use the lib/umdl code in umdl1 to reduce differences between umdl1 and umdl2 and thus slowly migrate to the new code. One thing I am unsure about is the name umdl vs. update-master-directory-list. I know everybody calls it umdl but the long name gives at least a small idea what it might do. But let's just call it umdl2, should be good.

PR #112, opened 2015-07-23
If the host has no categories do not auto-disable it.

adrianreber:
Simple fix to not auto-disable hosts if only certain categories are scanned.


pypingou:
Fair enough :) :+1:

PR #114, opened 2015-07-29
Create client subpackage with report_mirror in it.

adrianreber:
The s3sync scripts wants to run report_mirror. This creates the appropriate package.

### DIFF
--- a/utility/mirrormanager2.spec
+++ b/utility/mirrormanager2.spec
@@ -124,6 +124,16 @@ Requires(pre):  shadow-utils
 Install a number of utility scripts to be used manually or in cron jobs to
 run MirrorManager.
 
+%package client
+Summary:        Fedora mirror management system downstream mirror tools
+Group:          Applications/Internet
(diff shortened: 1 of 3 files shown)


ralphbean:
Sure, :+1:!

utility/mirrormanager2.spec
```diff
@@ -124,6 +124,16 @@ Requires(pre):  shadow-utils
 Install a number of utility scripts to be used manually or in cron jobs to
 run MirrorManager.
 
+%package client
+Summary:        Fedora mirror management system downstream mirror tools
+Group:          Applications/Internet
```
pypingou: No need for Group anymore

adrianreber: All other subpackages also have the Group information. I can remove them all.

pypingou: thinking more about it, might be still needed if we want to target EL5. I don't know for the client but I doubt MM2 will run on EL5.

PR #117, opened 2015-08-27
Sort metalinks alternates by timestamps descending

### LINKED ISSUE #116: Sort repomd.xml by time in metalink
It would help consumers of our metalink if the repomd.xml's are sorted by time, with the newest one at the top and the alternates being older in turn. 

See discussion in https://fedorahosted.org/fedora-infrastructure/ticket/4866

adrianreber:
This change is running in staging and can be tested with:

curl 'https://mirrors.stg.fedoraproject.org/metalink?repo=rawhide&arch=x86_64'

This resolves #116

### DIFF
--- a/mirrorlist/mirrorlist_server.py
+++ b/mirrorlist/mirrorlist_server.py
@@ -22,6 +22,7 @@ import sys
 from string import zfill, atoi
 import time
 import traceback
+from operator import itemgetter
 
 try:
     import threading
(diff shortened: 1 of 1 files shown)


mirrorlist/mirrorlist_server.py
```diff
@@ -161,7 +162,7 @@ def metalink(cache, directory, file, hosts_and_urls):
     preference = 100
     try:
         fdc = file_details_cache[directory]
-        detailslist = fdc[file]
+        detailslist = sorted(fdc[file], key=itemgetter('timestamp'), reverse=True)
```
pypingou: While this works, I wonder if we could do the ordering sooner, maybe directly in the pickle so that there is no speed impact on the mirrorlist. What do you think?

adrianreber: Good point. Although this list is usually not longer than 3 or 4 entries it would need to be sorted on every metalink request. Sorting it during pickle creation is definitely the right thing to do. I will submit a fixed PR.

adrianreber:
Closing as the list should be sorted during pickle creation and not during metalink creation.

PR #118, opened 2015-08-27
Add order option to get_file_detail()

adrianreber:
Until now get_file_detail() was always used with all optional
parameters specified and thus always the correct FileDetails were
returned. If only the Directory.ID and the name of the file are
specified this function returns the older entry (if multiple entries for
the wanted file exist). This adds an option to reverse the order and so
the newest entry about the given file should be returned.


pypingou:
:+1: nice and clean

PR #122, opened 2015-09-02
Add code to graph repomd.xml propagation

adrianreber:
The crawler has been extended to only retrieve the repomd.xml file and print out its sha256sum. To create diagrams of this output an additional script has been added. The script to create the diagrams has a few hard-coded assumptions coded into it which is not perfect but integrating it into mirrormanager2's repository makes sure the script is not lost.


pypingou:
Looks sane to me

Issue #125, opened 2015-09-03
Remove trailing slashes from URLs


adrianreber:
MM2 basically requires that all URLs do not have a trailing slash. There are some places in the code which check if there is a trailing slash or not but at least the metalink generation does not verify the URL which lead to:

 https://fedorahosted.org/fedora-infrastructure/ticket/4881

If all slashes would be removed when an URL is added it should be resolved.

PR #126, opened 2015-09-04
Remove trailing slash of site and host_category_url

pypingou:
Fixes https://github.com/fedora-infra/mirrormanager2/issues/125


adrianreber:
Looks good. Thanks. The site.org_url probably needs no '/' removed but for consistency it makes sense. Let's hope users do not start to add multiple slashes ;-).

pypingou:
Yeah, let's assume they'll be nice and deal with it if they don't :)

Issue #115, opened 2015-08-11
Host Category URLs were not correctly removed from the database


adrianreber:
See https://fedorahosted.org/fedora-infrastructure/ticket/4850

Changes to the Host Categories and corresponding URLs left MM in a state where the URLs were removed from the Host but still present in the database. Thus the same URL cannot be added again anywhere in MM.

adrianreber:
https://fedorahosted.org/mirrormanager/ticket/61

adrianreber:
Should be fixed.

PR #127, opened 2015-09-07
Deleting URLs if site or host is removed.

adrianreber:
Deleting a mirror site and/or host did not correctly delete the URLs:

https://github.com/fedora-infra/mirrormanager2/issues/115
https://fedorahosted.org/mirrormanager/ticket/61

As URLs can only exist once in the database the URLs with the deleted
site and/or hosts are still in the database and cannot be added to
another host or deleted. Right now these left-over URLs have to be
deleted manually.


pypingou:
:+1: thanks!

PR #130, opened 2015-09-11
Import generate-worldmap from MM1

adrianreber:
I cherry-picked all generate-worldmap related commits from the MM1 repository and then adapted the script to MM2. The results of the new mm2_generate-worldmap can bee seen at:

https://admin.stg.fedoraproject.org/mirrormanager/statistics/maps/mirrors.html

https://adrian.fedorapeople.org/map.png


ralphbean:
Looks good to me :)  :+1:

pypingou:
I was wondering if just the fact that we make use of data provided by a package license under GPL make our code GPL as well.

Also, I didn't check carefully the code, but do we block embargoed country?

Finally, the png you link to work but not the link to stg, is that expected?

adrianreber:
The link to stg is kind of wrong. I tested something manually and it was overwritten by ansible. Another version of the  map is here:

https://adrian.fedorapeople.org/map/mirrors.html

Concerning the license. I am not sure. The code (according to the comment) is based on examples from basemap which is licensed under GPLv2. So this seems to be the reason why this code has also been licensed under GPLv2. As I have cherry-picked all the related commits from MM1 the license was also included.

Embargoed countries are indeed blocked (line 63):

```
            if gir['country_code'] in embargoed_countries:
                print "skipping " + hn
                continue
```

mdomsch:
Licensing, and pulling in the large basemaps as dependencies, is why I
broke map generation out into a separate subpackage in the RPM.  I didn't
want all of MM to be licensed GPLv2 only because the map library being used
very infrequently (e.g. never in the last few years) is GPLv2.  We could
just as easily write generate_worldmap to collect the data and export it
into a CSV file, and then have a second program import the CSV file and
plot the map.

MMv1 did not block embargoed countries.  We had the ability should Fedora
Legal have had a requirement for us to do so, but were never given that
requirement to implement in MMv1.  Instead, our mirrors, to be mirrors,
must accept that _they_ prevent export to embargoed countries. [1]



[1]  https://fedoraproject.org/wiki/Embargoed_nations

pypingou:
If we don't block, I think we were asked once to at least not show them in the map ;-)

adrianreber:
The code excludes mirrors in embargoed countries and they are indeed not shown on the map. So that part of the PR should be good.

About the license, as it is based on @mdomsch's code its his call if this is GPLv2 or not. The comment says:

```
# Based on examples from python-GeoIP and python-basemap-examples
# Licensed under the GNU General Public License v2
```

I can also see comments in the basemap examples which are identical to the ones in the generate-worldmap script. So I would say, that the embargoed countries handling is correct and the license is also correct.

mdomsch:
Yes, that was cribbed from python-basemap-examples, and only trivially
edited.  Hence the simple generate-worldmap program is GPLv2.  That, and to
avoid installing python-basemap everwhere (it's really large), is why it's
in a separate RPM subpackage in MMv1.

adrianreber:
Ping, any more comments? As far as I can tell all open points have been resolved.

ralphbean:
:+1:

PR #132, opened 2015-09-22
Really ensure that new host is admin_active

spartacus06:
f59f7455 has a bug, setting host.admin_active = True before
form.populate_obj() which immediately sets it to false again
for non-admin users.

Switch those lines.


pypingou:
wfm, thanks for the fix!

PR #135, opened 2015-09-29
Cascade deletion and bring back the rsyncFilter endpoint

pypingou:


pypingou:
I did some quick testing but it could use some more

@adrianreber if you know what the output should look like :)
I'd be in to add unit-tests as well.

adrianreber:
The 'Cascade the deletion of the site_admin...' look like an obvious and correct change.

I have not ever used the rsync filter interface and it seems nobody else is using it as it is now gone for almost half a year without anybody complaining. It definitely seems like a very useful interface and thanks for bringing it back. I am :+1:  for merging.

Issue #131, opened 2015-09-11
HTTP crawl skips most files


adrianreber:
@pypingou your commit ced912addae7e407536a4696a6d9676eb0135d1e broke HTTP crawling. Most directories are now skipped. I have seen this now on multiple hosts but was only able to track it down today. This needs to be reverted and the problem which this tried to fix needs to be fixed in some other way.

An up to date host which does only provide HTTP for scanning has only the following directories listed as up to date. Only repodata directories and no directories with actual content.

    4/i386/debug/repodata
    4/i386/repodata
    4/ppc/debug/repodata
    4/ppc/repodata
    4/SRPMS/repodata
    4/x86_64/debug/repodata
    4/x86_64/repodata
    5/i386/debug/repodata
    5/i386/repodata
    5/ppc/debug/repodata
    5/ppc/repodata
    5/SRPMS/repodata
    5/x86_64/debug/repodata
    5/x86_64/repodata
    6/i386/debug/repodata
    6/i386/repodata
    6/ppc64/debug/repodata
    6/ppc64/repodata
    6/SRPMS/repodata
    6/x86_64/debug/repodata
    6/x86_64/repodata
    7/ppc64/debug/repodata
    7/ppc64/repodata
    7/SRPMS/repodata
    7/x86_64/debug/repodata
    7/x86_64/repodata
    testing/4/i386/debug/repodata
    testing/4/i386/repodata
    testing/4/ppc/debug/repodata
    testing/4/ppc/repodata
    testing/4/SRPMS/repodata
    testing/4/x86_64/debug/repodata
    testing/4/x86_64/repodata
    testing/5/i386/debug/repodata
    testing/5/i386/repodata
    testing/5/ppc/repodata
    testing/5/SRPMS/repodata
    testing/5/x86_64/debug/repodata
    testing/5/x86_64/repodata
    testing/6/i386/debug/repodata
    testing/6/i386/repodata
    testing/6/ppc64/debug/repodata
    testing/6/ppc64/repodata
    testing/6/SRPMS/repodata
    testing/6/x86_64/debug/repodata
    testing/6/x86_64/repodata
    testing/7/ppc64/debug/repodata
    testing/7/x86_64/debug/repodata
    testing/7/x86_64/repodata

pypingou:
If this specific commit broke something then we have a larger problem as before this commit we were trying to use a non-existing dict :-/

adrianreber:
[...]

PR #107, opened 2015-07-16
Fix the crawler to work with MM2

pypingou:

PR #138, opened 2015-10-12
But it is a dict

adrianreber:
There was some confusion about the list of files which is checked by the
crawler. The original code was looping over the keys of a dict. Which
seemed to been an unlikely result from a database query. The PR

https://github.com/fedora-infra/mirrormanager2/pull/107

changed it to loop over an existing database query result list.
This change, however, resulted that the crawler was only looking at
repodata directories:

https://github.com/fedora-infra/mirrormanager2/issues/131

The reason the crawler actually has to loop over the keys of a dict is
that umdl reads all the directories and creates a dict of the (maybe 10)
newest files in that directory. This dict is then pickled and stored in
the database and read by the crawler.

Instead of simply reverting the commit which removed the loop over the
dict, this change keeps all other improvements and only changes the loop
to use the pickled dict again.

Successfully tested in the staging environment.

### DIFF
--- a/utility/mm2_crawler
+++ b/utility/mm2_crawler
@@ -551,7 +551,7 @@ def check_ftp_file(hoststate, url, filedata, readable):
         if filedata is None:
             # The file/directory seems to exist
             return True
-        if float(line[4]) == float(filedata.size):
+        if float(line[4]) == float(filedata['size']):
             return True
     return False
 
@@ -628,7 +628,7 @@ def check_head(hoststate, url, filedata, recursion, readable, retry=0):
             # The file/directory seems to exist
             return True
         # fixme should check last_modified too
-        if float(filedata.size) == float(content_length) \
+        if float(filedata['size']) == float(content_length) \
                 or content_length is None:
             # handle no content-length header, streaming/chunked return
             # or zero-length file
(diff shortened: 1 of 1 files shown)


utility/mm2_crawler
```diff
@@ -988,10 +997,15 @@ def try_per_dir(d, hoststate, url):
...
+    # d.files is a dict which contains the last (maybe 10) files
+    # of the current directory. umdl copies the pickled dict
+    # into the database. It is either a dict or nothing.
+    if not isinstance(d.files, dict):
+        return False
+    for filename in d.files.keys():
         try:
-            if float(results[filename]['size']) != float(filedetail.size):
+            if (float(results[filename]['size']) !=
+                float(d.files[filename]['size'])):
```
pypingou: This would need an extra indentation

utility/mm2_crawler
```diff
@@ -804,15 +804,19 @@ def try_per_file(d, hoststate, url):
...
         return None
     exists = None
-    for filedetail in d.fileDetails:
-        filename = filedetail.filename
+    # d.files is a dict which contains the last (maybe 10) files
+    # of the current directory. umdl copies the pickled dict
+    # into the database. It is either a dict or nothing.
+    if not isinstance(d.files, dict):
+        return None
+    for filename in d.files.keys():
```
pypingou: You can drop the ``.keys()``, browsing a dict is much faster without it actually

adrianreber:
Thanks for the review. I updated the patch to include the necessary fixes.

pypingou:
:+1: for me

PR #146, opened 2015-11-14
Optionally display mirrorlist statistics

adrianreber:
The commits in this PR are adding additional tabs to the web frontend to display mirrorlist statistics and the map of all public mirrors. The results of this PR can currently be seen in staging:

https://admin.stg.fedoraproject.org/mirrormanager/statistics

https://admin.stg.fedoraproject.org/mirrormanager/maps

The backend code generating the statistics and maps is from MirrorManager1 and only the code to display the results is new code.


mirrorlist/mirrorlist_statistics.py
```diff
@@ -21,18 +21,18 @@
...
+import pylab
+import time
+import getopt
 
 start = time.clock()
 
 logfile = None
 dest = None
 offset = 0
+embargoed_countries = ['CU', 'IR', 'KP', 'SD', 'SY']
```
pypingou: Should we keep this in the configuration file?

mirrormanager2/app.py
```diff
@@ -1361,8 +1364,74 @@ def shutdown_session(exception=None):
...
+
+
+def check_for_statistics(date, cat):
+    try:
+        stat_file = APP.config['STATISTICS_BASE']
+        stat_file = os.path.join(
+            stat_file,
+            statistics_file_name(date, cat, 'txt')
+        )
+        os.stat(stat_file)
```
pypingou: What does this one do?  It doesn't seem like we're checking its output nor that it changes anything.

mirrormanager2/templates/fedora/maps.html
```diff
@@ -0,0 +1,12 @@
+{% extends "master_noauth.html" %}
+
+{% block title %}Maps{% endblock %}
+{%block tag %}maps{% endblock %}
+
+{% block content %}
+<h1>Maps</h1>
+{%- if config['SHOW_MAPS'] -%}
```
pypingou: Should we add an 'else' message ?

mirrormanager2/templates/fedora/statistics.html
```diff
@@ -0,0 +1,25 @@
...
+{% block title %}Statistics{% endblock %}
+{%block tag %}statistics{% endblock %}
+
+{% block content %}
+<h1>Statistics</h1>
+<p class="centered">Mirrorlist access breakdown by
+<a href="{{url_for('statistics')}}/{{ today }}/countries">country</a>
+| <a href="{{url_for('statistics')}}/{{ today }}/archs">architecture</a>
+| <a href="{{url_for('statistics')}}/{{ today }}/repositories">repository</a>
+{{today}}</p>
```
pypingou: I'd be in favor of adding a little text before the date, otherwise it kind of looks like ``repository <date>``, maybe ``on <date>``?

mirrorlist/mirrorlist_statistics.py
adrianreber: Hmm... Would make sense. Right now the script does not requite any MirrorManager infrastructure (configuration files, database access, imports). Which is nice as it can run anywhere as any user. If I start reading configuration files, which require special privileges, I could easily also connect to the database and read the list from there, which would make even more sense.

So either it will become less flexible by requiring certain hosts and users where and which it has to run, or it will stay the way it is now, but we have to maintain this list in the source code in addition to the list in the database. I have no preferences.

pypingou: The configuration file might be a good compromise, especially considering the config file could be trimmed down to only contain the information of interest.
It's just that from experience it's easier to adjust a configuration file than doing a new release or using the hotfix mechanism when we need to adjust this kinds of thing.

Note: do we store in the DB the country embargoed? I'm not even sure it's there.

mirrormanager2/app.py
adrianreber: Yes, this is a bit strange. The function assumes some things about its environment, which could be made clearer. The function basically wants to know if the file exists or not. If the stat() call fails an exception is thrown and the calling function gets a None back.

The function assumes about its environment that it is set up that if the stat() call works the file actually can also be read (which must not be true, but it is assumed). So, yes this could be better. I will change it to actually check if the file actually can be read.

mirrorlist/mirrorlist_statistics.py
adrianreber: Yes, changing the code for a configuration seems wrong. I will move it to the configuration file.

And yes, the DB has a table of embargoed countries, which is also used for the maps. It is only used by the map script so far, maybe it could also be removed from the DB and live in the configuration file only.

mirrormanager2/templates/fedora/maps.html
adrianreber: Yes :smile:

mirrorlist/mirrorlist_statistics.py
pypingou: Sounds good to me!

mirrormanager2/templates/fedora/statistics.html
adrianreber: Good idea.

mirrormanager2/templates/fedora/maps.html
```diff
@@ -0,0 +1,14 @@
...
+Map of all public mirrors - <a href="{{url_for('index')}}{{ config['INTERACTIVE_MAP'] }}">Interactive Map</a>
+<img src="{{url_for('index')}}{{ config['STATIC_MAP'] }}" border="0" alt="alt" class="centered"/>
+{%- else -%}
+Actually, this is disabled and you should not be here. Go away!
```
pypingou: lol

mirrormanager2/templates/fedora/master_noauth.html
```diff
@@ -28,6 +28,12 @@ <h4 class="hide">Main Site Links:</h4>
           <ul>
             <li id="homeTab"><a href="{{url_for('index')}}">Home</a></li>
             <li id="mirrorsTab"><a href="{{url_for('list_mirrors')}}">Mirrors</a></li>
+            {% if config['SHOW_STATISTICS'] %}
+            <li id="statsTab"><a href="{{url_for('statistics')}}">Statistics</a></li>
+            {% endif %}
+            {% if config['SHOW_MAPS'] %}
+            <li id="mapsTab"><a href="{{url_for('maps')}}">Maps</a></li>
+            {% endif %}
```
pypingou: Do we want the link on the auth master template as well?

adrianreber: The master template extends the "master_noauth.html" template, so that the links are always visible (if enabled).

pypingou:
Ok :+1: for me then

adrianreber:
Thanks for the review.

PR #147, opened 2015-11-27
mirrorlist: Remove --debug option

adrianreber:
The only use of the '--debug' option was to additionally log the
information for the mirrorlist statistics also to syslog. As the
mirrorlist statistics information is logged to its own file it does not
need to be duplicated to syslog. This completely removes the '--debug'
option.


pypingou:
:+1: for me :)

PR #151, opened 2015-12-15
More statistics changes and spec file integration

adrianreber:
https://admin.stg.fedoraproject.org/mirrormanager/propgation


mirrormanager2/app.py
```diff
@@ -1373,9 +1373,9 @@ def set_session():
 
 
 def statistics_file_name(date, cat, ext):
-    year = date.year
-    month = date.month
-    day = date.day
+    year = date.strftime('%Y')
+    month = date.strftime('%m')
+    day = date.strftime('%d')
```
pypingou: This kinda surprises me, what changed?

mirrormanager2/app.py
```diff
@@ -1436,3 +1436,39 @@ def statistics(date=None, cat='countries'):
...
+    """
+
+    prefix = os.path.basename(prefix)
+    # Right now propagation statistics are only generated
+    # for the 'development' prefix and 'fxx_updates' prefix.
+    if not prefix.startswith('f') and not prefix.startswith('d'):
+        prefix = 'development'
+
+    prop_base = APP.config['PROPAGATION_BASE']
+    stat_file = os.path.join(prop_base, prefix + "-repomd-propagation.svg")
```
pypingou: What about having this run by ``werkzeug.secure_filename()`` to make sure it's safe?

mirrormanager2/app.py
adrianreber: Oh, the files have zero padded dates like 2015-12-05 and .day returns an integer. So that breaks. The output file is generated with strptime('%Y-%m-%d') and is thus zero padded.

utility/mirrormanager2.spec
```diff
@@ -134,6 +134,24 @@ Client-side, run on each downstream mirror, to report back to the
 MirrorManager database a description of the content carried by that
 mirror.
 
+%package statistics
+Summary:        Scripts to generate MirrorManager statistics
```
pypingou: Do we have to adjust the license of the main package btw?

utility/mm2_propagation
```diff
@@ -166,42 +251,59 @@ fd.close()
...
+    "%-s-%s-repomd-propagation.pdf" %
+    (args.prefix, labels[len(labels) - 1].split('\n')[0]),
+    bbox_inches='tight')
+
+plt.savefig(
+    "%s-%s-repomd-propagation.svg" %
+    (args.prefix, labels[len(labels) - 1].split('\n')[0]),
+    bbox_inches='tight')
+
+print("%s-%s-repomd-propagation.pdf" %
```
pypingou: This is meant to be run in a cron job right? If so, should we only print when debugging?

mirrormanager2/app.py
pypingou: ok

adrianreber:
I updated the license, removed the debug output (if not especially requested) and the filename is now run through werkzeug.secure_filename(). Thanks for the review.

pypingou:
:+1: for me, do you want to patch the ansible playbook as well?

PR #152, opened 2015-12-16
setup_continents wants to use info we just loaded from the pkl

puiterwijk:


adrianreber:
Thanks for the fix. I verified this fix in staging and it works. :+1:

PR #153, opened 2015-12-20
statistics: Fix various matplotlib problems

adrianreber:
The matplotlib dependent scripts are trying to start a GTK interface if
no configuration file is present (like on my system) which tells that
the non-graphical interface should be used as default. With this it is
not explicitly stated in the code of the scripts that the non-graphical
version is wanted. In addition the dependency to draw the fonts with
Latex has been removed. Also switching to embargoed countries from
configuration file.


pypingou:
Looks sane to me

PR #156, opened 2016-01-11
Defer loading of host.config

adrianreber:
The field host.config is never used but increases the required memory dramatically for certain database operation. Its loading is now deferred and as it is never read the call write the data has also been removed.


adrianreber:
The MM1 code also does not store the config anywhere. So this change in behaviour has somehow crept into MM2. This is probably a database field which could be removed. Never read, never written.

puiterwijk:
Looks sane to me. :+1:

pypingou:
:+1: for me as well

adrianreber:
Thanks!

PR #157, opened 2016-01-14
Few small fixes to the emergency script

adrianreber:
I just tried the new emergency script in staging which was introduced with #154.

This PR has a few small fixes to make the script actually run and I also added it to the backend RPM as the script was not installed.

In addition to the described problems it seems not to be doing what it should do. For only one directory the script claims to have changed something:

```
[2016-01-14 08:50:36,228][       mm2    INFO] Clearing for repo pub/fedora/linux/atomic/23
[2016-01-14 08:50:36,236][       mm2    INFO] Files kept: summary
[2016-01-14 08:50:36,236][       mm2    INFO] Older versions deleted: summary
```

Everything else seems unchanged. The metalink output is also the same. The existing alternate entry is not removed. On each run the script produces the same output.

https://adrian.fedorapeople.org/mm/emergency-expire-repo-has-run
https://adrian.fedorapeople.org/mm/emergency-expire-repo-has-not-run

@puiterwijk

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,6 +57,7 @@ setup(
         'client/report_mirror',
         'mirrorlist/mirrorlist_statistics',
         'utility/mm2_crawler',
+        'utility/mm2_emergency-expire-repo',
         'utility/mm2_generate-worldmap',
         'utility/mm2_get_global_netblocks',
         'utility/mm2_get_internet2_netblocks',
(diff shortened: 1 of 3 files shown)


pypingou:
:+1: for me

@puiterwijk  do you want to review it as well?

puiterwijk:
This patch :+1:. Will look at fixing the script up, thanks!

PR #158, opened 2016-01-14
Finish the emergency cleanup script

puiterwijk:
It was not traversing subdirectories, and as such would not
reach $repo/repodata and the other important directories.

### DIFF
--- a/mirrormanager2/lib/model.py
+++ b/mirrormanager2/lib/model.py
@@ -27,6 +27,7 @@ import datetime
 import collections
 import logging
 import time
+import os
 
 import sqlalchemy as sa
 from sqlalchemy import create_engine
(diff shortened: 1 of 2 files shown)


mirrormanager2/lib/model.py
```diff
@@ -27,6 +27,7 @@
 import collections
 import logging
 import time
+import os
```
pypingou: Left over from debugging?

mirrormanager2/lib/model.py
```diff
@@ -827,25 +828,29 @@ def emergency_expire_old_file_details(self, session):
         CAUTION: this will make all slightly outdated mirrors be non-trusted,
         so should be used sparingly!
         """
+        if not self.directory:
+            return False
 
-        files = session.query(FileDetail). \
-            filter_by(directory_id=self.directory_id). \
-            order_by(FileDetail.timestamp.desc()).all()
+        subdirs = session.query(Directory). \
```
pypingou: I kinda like more the style where the braces are used to avoid using ``\``, something like:

````
session.query(
    Directory
).filter(
   ...
).all()
````
Same for below

But this is really just style

mirrormanager2/lib/model.py
pypingou: NM I see where it's being used

pypingou:
One comment on style, otherwise looks good

adrianreber:
Also see #157

adrianreber:
I just verified the latest code in staging with Fedora 22 and can confirm this works as expected. Thanks! I am with @pypingou concerning the style as it is pretty consistent right now in the codebase and it would be nice to use the same style also here.

pypingou:
:+1: for me

adrianreber:
also :+1: here

PR #169, opened 2016-05-25
Propagation: Added path to propagation results

adrianreber:
This change is required to make the propagation graphing actually work.
It seems this has been fixed manually on the production systems.


adrianreber:
Unfortunately I only fixed this on the production system directly a few weeks/months ago without bringing the change to git and the RPMs.

pypingou:
Simple as well, :+1:

adrianreber:
Thanks

PR #171, opened 2016-06-02
umdl: mark paths containing /stage/ also as testing

adrianreber:
As described in

https://fedorahosted.org/fedora-infrastructure/ticket/5185

umdl picks up repositories and versions in the alt/ tree in the staging
directory, which should not be presented on the MirrorManager main page.

umdl already has a check to see if a version is a test version and this
is now extended to also mark releases/repositories/versions in paths
containing 'stage' as 'isTest'.


pypingou:
:+1: for me

ralphbean:
:+1: here too!

Issue #100, opened 2015-06-22
Add a way to specify you want only https urls from metalink


nirik:
Some folks want all their traffic to use ssl, we should offer a option to the metalink url that makes it return just https using mirrors. Something like ?method=https or the like in the url. 

Along with this we might consider mailing mirror admins and asking if they would update to https where they have https available. 

Once enough mirrors offered https we could make it the default perhaps.

adrianreber:
This is now running on the Fedora production systems.

mdomsch:
Glad to see this, but the whole point of metalinks is to let the clients
decide what they want to use - give them all the possible URLs and then
they choose what they want. This properly belongs in a dnf configuration,
not MM2. As it stands, the user now has to change the dnf configuration
anyhow, to add the specifier on the metalink URL.

adrianreber:
Yes, I agree with @mdomsch. The official way to make metalink clients use HTTPS must be different. This needs some additional work on other parts (could be dnf). At least there is now a way for people who want to use it and we also know that it works and it is a starting point. It was also a good way to get more HTTPS URLs in the database. But, yes, now that more HTTPS URLs are in the database an option to make dnf prefer HTTPS URLs if available in the metalink sounds like a good idea.

nirik:
Sure, that makes some sense... I can ask dnf / librepo maintainers about options there.

nirik:
And there is already https://bugzilla.redhat.com/show_bug.cgi?id=1229050 (dnf should provide a protocol option) and https://bugzilla.redhat.com/show_bug.cgi?id=1273051 (librepo should provide a supported protocol list).

PR #172, opened 2016-06-05
mirrorlist: add protocol option

adrianreber:
This adds the option specify the desired protocol of the mirrorlist/metalink as described in #100 

The code is currently running in staging and the result can be seen at:

http://mirrors.stg.fedoraproject.org/mirrorlist?repo=rawhide&arch=x86_64&country=jp&protocol=rsync
http://mirrors.stg.fedoraproject.org/mirrorlist?repo=rawhide&arch=x86_64&country=jp&protocol=ftp
http://mirrors.stg.fedoraproject.org/metalink?repo=rawhide&arch=x86_64&country=gb&protocol=https
http://mirrors.stg.fedoraproject.org/metalink?repo=rawhide&arch=x86_64&country=gb&protocol=rsync

### DIFF
--- a/mirrorlist/mirrorlist_client.wsgi
+++ b/mirrorlist/mirrorlist_client.wsgi
@@ -74,7 +74,7 @@ def real_client_ip(xforwardedfor):
 def request_setup(environ, request):
     fields = [
         'repo', 'arch', 'country', 'path', 'netblock', 'location',
-        'version', 'cc'
+        'version', 'cc', 'protocol'
     ]
     d = {}
     request_data = request.GET
--- a/mirrorlist/mirrorlist_client.wsgi
+++ b/mirrorlist/mirrorlist_client.wsgi
@@ -165,7 +165,7 @@ def application(environ, start_response):
         text = ""
         text += message + '\n'
         for (hostid, url) in results:
-            text += url + '\n'
+            text += url[0] + '\n'
         results = text
         response.headers['Content-Type'] = "text/plain"
     elif resulttype == 'metalink':
--- a/mirrorlist/mirrorlist_server.py
+++ b/mirrorlist/mirrorlist_server.py
@@ -409,7 +409,7 @@ def trim_to_preferred_protocols(hosts_and_urls):
 
         for p in try_protocols:
             if p in protocols:
-                url = protocols[p]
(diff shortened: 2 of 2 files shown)


mirrorlist/mirrorlist_client.wsgi
```diff
@@ -165,7 +165,7 @@ def application(environ, start_response):
         text = ""
         text += message + '\n'
         for (hostid, url) in results:
-            text += url + '\n'
+            text += url[0] + '\n'
```
pypingou: This change surprises me a little

adrianreber: This is described in the first commit. To trim the protocols to only contain the requested protocols I am re-using the function trim_to_preferred_protocols(). This function is currently called after the metalink is created and changes the result from [(hostid, [url, url, url]), ...] to [(hostid, url), ...], which was okay for the mirrorlist back-end as it only returned one URL per host. As the function trim_to_preferred_protocols() already reduces the URLs returned per host to a defined set (https, http, ftp) for metalink I wanted to re-use this function to trim the URL list to the requested protocol.

To be able to re-use trim_to_preferred_protocols() it needs to return a result which metalink and mirrorlist can handle. metalink wants multiple URLs per host and therefore I am now always returning [(hostid, [url, url, url]), ...] and [url, url, url] now always contains exactly one element for mirrorlist and therefore url[0].

pypingou: ok thanks

pypingou:
:+1: for me

nirik:
Looks great! as a aside, we should likely ask mirrors to add more https links before we advertise this feaure. Or perhaps even scan all mirrors with http:// links and see how many support https and auto add those.

adrianreber:
Yes, scanning for https URLs should be possible and also easier than the ftp URLs removal.

adrianreber:
I scanned all http URLs and replaced 'http' with 'https' and added the result to the database. This increased the number of https URLs from 24 to 112. We have 920 http URLs in the database.

The metalink will start to redirect users to http and https URLs so people will be hitting more https mirrors in about 30 minutes.

I have not verified that the content provided via the same https URL is the same as the content from the http URL.

PR #175, opened 2016-06-19
Add logging to checkin code

puiterwijk:
This should help with finding issues with checkin.

### DIFF
--- a/mirrormanager2/lib/hostconfig.py
+++ b/mirrormanager2/lib/hostconfig.py
@@ -25,9 +25,14 @@ MirrorManager2 Host configuration.
 
 import mirrormanager2.lib
 
+import logging
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
 def validate_config(config):
     message = ''
     if type(config) != dict:
+        logging.critical("NON-DICT SUBMITTED: %s" % config)
         message += 'config file is not a dict.\n'\
         'Please update your copy of report_mirror.\n'
         return (False, message)
--- a/mirrormanager2/xmlrpc.py
+++ b/mirrormanager2/xmlrpc.py
@@ -27,6 +27,7 @@ import base64
 import pickle
 import json
 import bz2
+import logging
 
 import flask
 from flaskext.xmlrpc import XMLRPCHandler, Fault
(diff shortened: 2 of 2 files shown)


mirrormanager2/lib/hostconfig.py
```diff
@@ -25,9 +25,14 @@
 
 import mirrormanager2.lib
 
+import logging
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
```
pypingou: I'm not entirely sure if we should use ``__name__`` here vs say ``mirrormanager``

pypingou:
:+1: on the idea, just wondering about the name of the logger

mirrormanager2/lib/hostconfig.py
puiterwijk: Sure, I just used this since it's the only thing going into the httpd logs, so we can easily find which part of the code wrote this log.

pypingou: if it doesn't matter, I think I'd rather have it be ``mirrormanager`` as not everyone may be running MM on a dedicated host

puiterwijk: How about mirrormanager.lib.hostconfig?

bowlofeggs: I like the use of ```__name__``` as it gives a nice python path to this module, which is handy for quickly identifying where logging statements came from.

pypingou: But it makes it harder to configure the log (and especially the logging level)

bowlofeggs: I believe that you can set the root logger's log level, and all child loggers will inherit it. Something like this should do it:

```
import logging

root_logger = logging.getLogger()
root_logger.setLevel(logging.INFO)
```

I can't remember whether it's important for that statement to happen earlier than all other statements, or if it can happen at any time. In Pulp there is a similar statement very early during process startup, and all the loggers inherit that logger's log level (and handler too). I wrote most of the code in Pulp's logging setup, and you can see it here:

https://github.com/pulp/pulp/blob/pulp-2.8.4-1/server/pulp/server/logs.py

Line 53 is where the root logger gets set up.

pypingou: Ok cool :)

Does that work fine as well when specifying the logging via a config file?

puiterwijk: So, this is exactly what I'm doing with logging.basicConfig:

>  logging.basicConfig([**kwargs])
>     Does basic configuration for the logging system by creating a StreamHandler with a default Formatter and adding it to the root logger.

adrianreber:
Is this ready to be merged? If yes, I will make a release tomorrow, ready to be installed after the freeze is over.

pypingou:
It has a :+1: for a while now :)

Issue #178, opened 2016-06-22
AttributeError: 'list' object has no attribute 'startswith'


adrianreber:
In staging:

Traceback (most recent call last):
  File "/usr/share/mirrormanager2/mirrorlist_client.wsgi", line 156, in application
    results = keep_only_http_results(results)
  File "/usr/share/mirrormanager2/mirrorlist_client.wsgi", line 129, in keep_only_http_results
    if url.startswith(u'http'):
AttributeError: 'list' object has no attribute 'startswith'

This request triggers the error: https://mirrors.stg.fedoraproject.org/mirrorlist?path=pub/fedora/linux/&redirect=1

Everything using redirect=1 is broken since 0.7.2 has been installed in staging.

adrianreber:
Fixed version deployed.

PR #179, opened 2016-06-22
mirrolist: fix &redirect=1

adrianreber:
The commits

  ed7e45903f381a07395b7960a719934414c7c8f5
  236532af7a46d50930bdc3f9c8386a063d5960cf

which were used to support the '&protocol=' option broke the mirrorlist
option to answer with a HTTP redirect (302). This fixes the redirect
code path to also work with the new data format returned by the
mirrorlist server.

### DIFF
--- a/mirrorlist/mirrorlist_client.wsgi
+++ b/mirrorlist/mirrorlist_client.wsgi
@@ -123,12 +123,15 @@ def request_setup(environ, request):
     return d
 
 
-def keep_only_http_results(input):
-    output = []
+def get_first_http_url(input):
+    """ Only used for the redirect case. In the case
+    a redirect has been requested only the first URL
+    is returned starting with 'http'."""
     for hostid, url in input:
-        if url.startswith(u'http'):
-            output.append((hostid, url))
-    return output
+        for u in url:
+            if u.startswith(u'http'):
+                return u
+    return None
 
 
 def application(environ, start_response):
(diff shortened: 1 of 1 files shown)


adrianreber:
Fixes #178

puiterwijk:
Looks sane to me.

mirrorlist/mirrorlist_client.wsgi
```diff
@@ -123,12 +123,14 @@ def request_setup(environ, request):
...
+def get_first_http_url(input):
+    """ Only used for the redirect case. In the case
+    a redirect has been requested only the first URL
+    is returned starting with 'http'."""
     for hostid, url in input:
-        if url.startswith(u'http'):
-            output.append((hostid, url))
-    return output
+        if url[0].startswith(u'http'):
+            return url[0]
```
pypingou: Do we want to check the other URLs if the first one doesn't start with http?

pypingou:
Do we want to plan unit-tests for these changes?

mirrorlist/mirrorlist_client.wsgi
adrianreber: I was also thinking about it and it is probably the right thing to do. I was not doing it as there usually always is a mirror with at least one HTTP URL. But now it sounds like a bad idea to rely on that assumption. I will update the PR.

adrianreber:
How would unit-tests look like for the mirrorlist server?

Generate a small pkl, start the mirrolist server and query all possible mirrorlist/metalink combinations?

pypingou:
oh shoot, missed that it was the mirrorlist, your approach seems right yes.

adrianreber:
@pypingou now with a loop over all URLs

pypingou:
Cool, :+1: for me :)

PR #182, opened 2016-06-27
Add json checkin endpoint

puiterwijk:
This is a simple json-only endpoint that does not require xmlrpc.
It accepts both plain json and gzip compress json as input.

Signed-off-by: Patrick Uiterwijk puiterwijk@redhat.com

### DIFF
--- a/mirrormanager2/checkin.py
+++ b/mirrormanager2/checkin.py
@@ -28,6 +28,7 @@ import pickle
 import json
 import bz2
 import logging
+import zlib
 
 import flask
 from flaskext.xmlrpc import XMLRPCHandler, Fault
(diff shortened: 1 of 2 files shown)


adrianreber:
Would it be possible to log the host id and the IP address from which the request was coming from?

jasontibbitts:
Dumb question, but with this endpoint, does the entire upload happen as part of the GET request?  Seems like this could be a lot of data, especially if we ever extend the amount of data being sent (to, say, include the original plan of sending the complete file listing with sizes).

jasontibbitts:
Not to be a pain, but I was just wondering if anything has happened here.  I'm probably not going to have much time to work on this until flock so there's no hurry.

mirrormanager2/checkin.py
```diff
@@ -40,6 +41,34 @@
 logger = logging.getLogger(__name__)
 
 
+@APP.route('/checkin/json', methods=['POST'])
+def checkin_json():
```
bowlofeggs: Please add a docblock here.

mirrormanager2/checkin.py
```diff
@@ -40,6 +41,34 @@
...
+    try:
+        if ctype == 'application/gzip':
+            data = json.loads(zlib.decompress(flask.request.data, 16+zlib.MAX_WBITS))
+        elif ctype == 'application/json':
+            data = json.loads(flask.request.data)
+        else:
+            return 'Invalid content type', 415
+    except Exception as ex:
+        logging.error('Unable to read input data: %s' % ex)
+        return 'Error decoding', 400
```
bowlofeggs: What Exceptions do you expect? It's often a nice pattern to handle specific expected Exceptions and then have a final catch-all like this, so you can provide nice error messages for the expected ones.

mirrormanager2/checkin.py
```diff
@@ -40,6 +41,34 @@
...
+
+    try:
+        r, message = read_host_config(SESSION, config)
+    except Exception as ex:
+        logging.error('Unable to parse input data: %s' % ex)
+        return 'Error parsing', 500
```
bowlofeggs: Would there be any expected Exceptions here too that we could catch and provide a nicer error message?

mirrormanager2/checkin.py
```diff
@@ -40,6 +41,34 @@
...
+
+    if r is not None:
+        logging.info('JSON checkin succesful: %s' % message)
+        return message
+    else:
+        logging.error('Error during JSON checkin: %s' % message)
+        return message, 406
```
bowlofeggs: I take it that ```read_host_config()``` doesn't always raise an Exception when there's a problem and that's why we have this?

adrianreber:
@puiterwijk are you still interested in this PR?

ryanlerch:
closing this one as stale -- no response from original PR poster.

please reopen and fix conflicts if still needed

PR #190, opened 2016-09-23
Skip disabled repositories in the overview

adrianreber:
MirrorManager still picks up old and archived versions of repositories.
This uses the existing disabled flag to ignore those repositories in the
overview.

See https://fedorahosted.org/fedora-infrastructure/ticket/4625

### DIFF
--- a/mirrormanager2/lib/model.py
+++ b/mirrormanager2/lib/model.py
@@ -769,6 +769,8 @@ class Version(BASE):
         '''
         arches = set()
         for repo in self.repositories:
+            if repo.disabled:
+                continue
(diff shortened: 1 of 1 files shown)


puiterwijk:
It looks good to me, but just broke production, so maybe let's verify what happened there first :)

adrianreber:
This change doesn't seem to be the direct cause for the broken mirrorlist in production. But setting the rawhide repository to disabled for a certain arch, disables all rawhide repositories.So this code change is not wrong but the database field 'disabled' has another meaning. Closing this PR as it cannot be used.

PR #191, opened 2016-09-26
add Vagrant setup for mirrormanager2

ryanlerch:


README.rst
```diff
@@ -17,6 +17,36 @@ For more details `mirrormanager <https://fedorahosted.org/mirrormanager/>`_
 Hacking
 -------
 
+Hacking with Vagrant
+~~~~~~~~~~~~~~~~~~~~
+Quickly start hacking on mirrormanager2 using the vagrant setup that is included
+in the repo is super simple.
+
+First, install Ansible, Vagrant and the vagrant-libvirt plugin from the official Fedora
```
puiterwijk: Also add vagrant-sshfs?

Vagrantfile
```diff
@@ -0,0 +1,17 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure(2) do |config|
+  config.vm.box_url = "https://download.fedoraproject.org/pub/fedora/linux/releases/23/Cloud/x86_64/Images/Fedora-Cloud-Base-Vagrant-23-20151030.x86_64.vagrant-libvirt.box"
+  config.vm.box = "f23-cloud-libvirt"
```
puiterwijk: Would it be an idea to use f24?

Vagrantfile
```diff
@@ -0,0 +1,17 @@
...
+  config.vm.network "forwarded_port", guest: 5000, host: 5001
```
puiterwijk: Should these be separate?

devel/ansible/playbook.yml
```diff
@@ -0,0 +1,8 @@
+---
+- hosts: all
+  become: true
+  become_method: sudo
+  vars:
```
puiterwijk: Empty vars?

devel/ansible/roles/core/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
...
+- name: Install basic packages
+  dnf:
+      name: "{{ item }}"
+      state: present
+  with_items:
+      - bash-completion
+      - dstat
+      - htop
+      - screen
+      - tmux
```
puiterwijk: Both screen and tmux?

README.rst
```diff
@@ -17,6 +17,40 @@ For more details `mirrormanager <https://fedorahosted.org/mirrormanager/>`_
...
 -------
 
+Hacking with Vagrant
+~~~~~~~~~~~~~~~~~~~~
+Quickly start hacking on mirrormanager2 using the vagrant setup that is included
+in the repo is super simple.
+
+First, make a copy of the Vagrantfile example::
+
+    $ cp Vagrantfile.example Vagrantfile
```
pypingou: If we don't edit the vagrantfile, what is the advantage of not shipping it as ``vagrantfile``?

ryanlerch: I think the idea here is to open it up if people do want to change the vagrant file, they can -- depending on if they want different network settings etc.

ryanlerch:
I think we also need to figure out a way to create some sample data here, or import a datadump from Prod (if its not too large that is) or a subset of it.

README.rst
pypingou: ok

pypingou:
The database in prod contains some confidential information and thus we can't use it for dev. Generating a subset of it would likely end up being as much work as generating some fake/sample data.

Issue #183, opened 2016-07-12
Mirrors with only a HTTPS URL cannot be crawled


adrianreber:
The function method_pref() in the crawler cannot correctly handle mirrors which only provide a HTTPS URL. The simplest fix would probably be to remove the ':' in `u.startswith('http:')`.

haydenjames:
Avoid adding https and use http instead? Or fixed?

adrianreber:
At least one HTTP or RSYNC URL needs to be available. If that is the case HTTPS can be added.

adrianreber:
Fixed by #192

PR #192, opened 2016-09-29
crawler: also crawl https-only mirrors

adrianreber:
The function which selects the crawl protocol only knows
about 'rsync:', 'http:' and 'ftp:'.

A mirror which only provides 'https' will never be crawled.

This only changes 'http:' to 'http'.

### DIFF
--- a/utility/mm2_crawler
+++ b/utility/mm2_crawler
@@ -748,7 +748,7 @@ def sync_hcds(session, host, host_category_dirs, repodata):
 
 def method_pref(urls, prev=""):
     """ return which of the hosts connection method should be used
-    rsync > http > ftp """
+    rsync > http(s) > ftp """
     pref = None
     for u in urls:
         if prev.startswith('rsync:'):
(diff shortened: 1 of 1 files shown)


utility/mm2_crawler
```diff
@@ -748,15 +748,15 @@ def sync_hcds(session, host, host_category_dirs, repodata):
...
-    rsync > http > ftp """
+    rsync > http(s) > ftp """
     pref = None
     for u in urls:
         if prev.startswith('rsync:'):
             break
         if u.startswith('rsync:'):
             return u
     for u in urls:
-        if u.startswith('http:'):
```
puiterwijk: I would prefer a "if u.startswith(('http:', 'https:'))", so that we have the colon included at least, so we know it's a protocol indication.

adrianreber: Thanks, changed.

adrianreber:
Fixes #183

PR #193, opened 2016-10-08
Remove <0.80 requirement on IPy

puiterwijk:
Signed-off-by: Patrick Uiterwijk puiterwijk@redhat.com

### DIFF
--- a/utility/mirrormanager2.spec
+++ b/utility/mirrormanager2.spec
@@ -21,7 +21,7 @@ BuildRequires:  python-flask-admin
 BuildRequires:  python-flask-xml-rpc
 BuildRequires:  python-flask-wtf
 BuildRequires:  python-wtforms
-BuildRequires:  python-IPy < 0.80
+BuildRequires:  python-IPy
 BuildRequires:  python-dns
 BuildRequires:  python-fedora >= 0.3.33
 BuildRequires:  python-fedora-flask >= 0.3.33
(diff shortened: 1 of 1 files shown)


pypingou:
Because the issue is fixed?

puiterwijk:
Because I have not been able to see any reason why it was added (requirements.txt doesn't list the version requirement, so pip would happily install the latest) and this has not been changed since its introduction two years ago in the spec file.

pypingou:
In theory I have nothing against, but I'd rather finding back why this was added before actually merging this PR.

bowlofeggs:
The IPy 0.80 release notes don't say anything that sounds scary to me:

https://github.com/autocracy/python-ipy/blob/master/ChangeLog

I'm comfortable with dropping the requirement, given the release notes. I agree with @pypingou that it would be helpful to know why that requirement was put in place, but it's also possible that that information has been lost.

adrianreber:
I cannot load the pkl on Fedora with python-IPy-0.81-16.fc25.noarch:

TypeError: ('__init__() takes at least 2 arguments (1 given)', <class 'IPy.IP'>, ())

Same code works on CentOS7 with python-IPy-0.75-6.el7.noarch

Not really sure where this comes from.

pypingou:
It is my understanding that this has now been done on the servers of the Fedora Infra, and afaict it all works.

So I guess I should lift my suspicions and say 👍 to this PR :)

adrianreber:
The main problem seems to be that it is not possible to mix versions. A pkl generated with an older version does not work in a newer version. But yes, let's merge this.

PR #196, opened 2016-11-16
mm2_move-devel-to-release: adapt to latest repository layout

### LINKED ISSUE #195: mm2_move-devel-to-release needs to be adapted for new repository layout
The mm2_move-devel-to-release script still assumes that the path releases/version/... includes the 'Everything/' part and the development/version/... does not include 'Everything/'. Thus it looks for wrong paths in the database and fails. This needs to be fixed (already vim-fixed in the Fedora production systems).

adrianreber:
mm2_move-devel-to-release was still assuming that the 'releases/'
directory has the 'Everything/' part in it but that 'Everything/' does not
exist under 'development/'. As this has changed this script needs to
adapt to the new layout. Without this change the script was looking for
directories like

pub/fedora-secondary/releases/24/Everything/Everything/ppc64/debug/tree

which do not exist.

Bug report: https://bugzilla.redhat.com/show_bug.cgi?id=1395695

Fixes #195

### DIFF
--- a/utility/mm2_move-devel-to-release
+++ b/utility/mm2_move-devel-to-release
@@ -49,7 +49,7 @@ def move_devel_repo(session, category, version):
         sys.exit(1)
 
     oldpattern = os.path.join('development', version)
-    newpattern = os.path.join('releases', version, 'Everything')
+    newpattern = os.path.join('releases', version)
     oldRe = re.compile(oldpattern)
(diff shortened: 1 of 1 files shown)


pypingou:
Looks good to me, I just wish we would stop constantly changing those paths :( Or at least that it was clearly announced somewhere.

Issue #149, opened 2015-12-14
Description (help_text) for add host category URL is missing


adrianreber:
Ticket https://fedorahosted.org/fedora-infrastructure/ticket/5016 made clear, that the old MM1 help text is also required for the form ['New Host Category URL'](../tree/master/mirrormanager2/templates/fedora/host_category_url_new.html). This private attribute has no relation to the notion of a private mirror and the idea behind it seems to be private URLs which can be used between mirrors to sync content.

It could be removed as the concept of private inter-mirror URLs is not used at all.

adrianreber:
Necessary help text added and private URLs are admin only now.

PR #198, opened 2017-01-31
Private URLs are now restricted to admins

adrianreber:
MirrorManager has the concept of private URLs which should be used
for inter-mirror syncing. Unfortunately they are sometimes confused
with the concept of private mirrors. Private URLs will never show up in
the mirrorlist/metalink and if a private mirror (only for a certain
network or host) marks a URL as private it will not show up.

As the private flag is not reflected anywhere on the web pages it can
only be seen in the database.

With this change a private URL is now also displayed as private in the
host category URL view and only admins can mark URLs as private.

### DIFF
--- a/mirrormanager2/app.py
+++ b/mirrormanager2/app.py
@@ -1153,6 +1153,7 @@ def host_category_url_new(host_id, hc_id):
     if form.validate_on_submit():
 
         host_category_u = model.HostCategoryUrl()
+        private = host_category_u.private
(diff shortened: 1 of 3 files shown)


adrianreber:
Trying to fix #149 as a mirror marked URLs as private and I was debugging it for hours.

pypingou:
Looks fine to me

adrianreber:
Thanks. Closes #149

Issue #194, opened 2016-10-18
Provide option to make a mirror globally available


adrianreber:
A new option to mark a mirror globally available would help in cases where not many mirrors are available like: https://pagure.io/fedora-infrastructure/issue/5528

If the master mirror would be marked as 'global' with a low priority there would always be a least one mirror which is up to date. Having a low priority local mirrors would be used before contacting the master mirror. Also, in case of important updates at least one mirror with the correct repomd.xml checksum would be available.

This could also help to use mirrors which are based on anycast as we currently have no way to use anycast mirrors with our crawler/continent/country setup.

adrianreber:
This option should be admin only.

adrianreber:
There is already support in the [mirrorlist server](https://github.com/fedora-infra/mirrormanager2/blob/master/mirrorlist/mirrorlist_server.py#L563) to extent the result with mirrors from the global list if less than 3 mirrors are returned.

I will increase this from 3 to 5 and make it configurable from the command-line if it needs to be changed in the future.

adrianreber:
PR #199 offers a work around.

An option to add certain mirrors to all mirrorlist/metalink results would still be helpful. This requires changes in the web-frontend, pkl creation and mirrorlist.

mdomsch:
In the past we've used a low bandwidth value, combined with the 'always_up2date' flag on the mirror category, to make sure the Fedora-owned public-facing servers appeared at the end of the mirrorlist as a mirror of last resort.

adrianreber:
@mdomsch are you sure? The mirrorlist_server.py code has not change much since you wrote it and I have not found the place where 'always_up2date' mirrors are added to the mirrorlist.

The master mirror is still marked as 'always_up2date' but I only see it on 'country=global' and 'country=us' mirrorlists.

mdomsch:
always_up2date flag is handled in the pickle creation queries.

adrianreber:
Thanks. I see it in the old and in the new code but the new code does not include it in the results.

RPM Fusion uses the old backend code (for pkl generation) and the new mirrorlist servers but the always_up2date flag is not used. I could test it there to see how it works.

Anyway, good to know that this is an additional meaning of the 'always_up2date' flag.

PR #199, opened 2017-02-01
mirrorlist: at least 5 mirrors should be returned for country/continent

adrianreber:
If less than three mirrors for a country/continent are found the
mirrorlist is extended by the global mirrorlist. This increases the
value from three to five as there have been problems especially with
countries/continents with a low number of mirrors (Australia).

In addition the minimum number of mirrors threshold can now be specified
on the command-line: -m, --minimum

Important to know is that if a specific country is requested from the
mirrorlist using '&country=AU' described mechanism does not take effect,
only for regular requests (without country) or requests for a specific
IP ('&ip=xxx.xxx.xxx.xxx').

### DIFF
--- a/mirrorlist/mirrorlist_server.py
+++ b/mirrorlist/mirrorlist_server.py
@@ -41,6 +41,9 @@ cachefile = '/var/lib/mirrormanager/mirrorlist_cache.pkl'
 internet2_netblocks_file = '/var/lib/mirrormanager/i2_netblocks.txt'
 global_netblocks_file = '/var/lib/mirrormanager/global_netblocks.txt'
 logfile = None
+# If not at least 'minimum' mirrors are found for a country/continent,
+# mirrors from the global list are appended to the country/continent list
+minimum = int(5)
 must_die = False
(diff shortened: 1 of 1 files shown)


adrianreber:
Possible solution for the problems described in #194

It still would make sense to actually implement what is described in #194

mirrorlist/mirrorlist_server.py
```diff
@@ -927,6 +931,11 @@ def parse_args():
                 logfile = open(argument, 'a')
             except:
                 logfile = None
+        if option in ("-m", "--minimum"):
+            minimum = int(argument)
```
pypingou: Does this need a ``global minimum`` above?

adrianreber: There is one in line 910

pypingou: Ok cool :)

pypingou: Ah, I see it now :)

pypingou:
Looks fine to me :) 👍

adrianreber:
Thanks

Issue #145, opened 2015-11-11
Text for "Master rsync server Access Control List IPs" misleading


adrianreber:
Each host has the possibility to define "Master rsync server Access Control List IPs" which has following description:

> These host DNS names and/or IP addresses will be allowed to rsync from the master rsync/ftp servers. List here all the machines that you use for pulling.

This is not true. Access to the master mirror in Fedora is not granted automatically. Maybe we should hide this part for Fedora's MirrorManager installation. There used to be a way for mirrors to query all hosts listed in this field via /rsync_acls/ to populate the "hosts allow" field of the local rsync daemon. This does not exist in MirrorManager2 any more as it never had many users.

So we could remove this functionality at least from the Fedora templates.

srisankethu:
@ralphbean Is this done? Or can I make changes?

adrianreber:
@srisankethu This issue is still open.

adrianreber:
See #200

PR #200, opened 2017-02-01
Remove 'Master rsync server Access Control List IPs' section

adrianreber:
This adds a configuration option to enable/disable the 'Master rsync
server Access Control List IPs' section in the host template. As Fedora
has never (or at least for a long time not) used this functionality this
disables if by default in the Fedora templates.

It could have been completely removed from the Fedora templates, but as
long as the Fedora templates are the only templates I just conditionally
disabled it.


adrianreber:
Fixes #145

pypingou:
👍 for me

PR #202, opened 2017-03-05
compress crawler logs in logrotate. They are small, but they add up

nirik:
compress crawler logs, save us some space on the crawler machines. 🐄


adrianreber:
Thanks.

Issue #71, opened 2015-05-14
Hide certain fields for non-admins


adrianreber:
MM1 was not displaying all possible fields to non-admin users. At least site->admin_active and host->admin_active should not be visible and changeable as well as Peer ASNs for example. The complete list of items to hide from non-admin users has to be looked up in the MM1 source code.

pypingou:
I have started on this but I only did the ``admin active`` fields

pypingou:
For the ``Peer ASNs`` do you mean the ``asn`` or ``asn_clients`` field?

adrianreber:
Neither, further down are different options like *Site-local Netblocks* and also *Peer ASNs* in the host view. *Peer ASNs* should only be added by admins. As well as *Site-local Netblocks* with a netmask larger than /16.

pypingou:
By admins, you mean mirrormanager admins or the admins of the site/host ?

adrianreber:
mirrormanager admins

pypingou:
We can easily hide the whole 'Site-local Netblocks' but hiding it for some netmasks might end-up being tricky.
Imagine a mirror admin makes a typo and replaces a /10 netmask with /16, they won't be able to fix the typo.

adrianreber:
Comparing it again to MM1. This is how MM1 did this:

https://git.fedorahosted.org/cgit/mirrormanager/tree/server/mirrormanager/controllers.py#n646

adrianreber:
MM1 has been moved to pagure

https://pagure.io/mirrormanager/blob/master/f/server/mirrormanager/controllers.py#_655

adrianreber:
PR #79 was partially solving this issue

adrianreber:
Should all be fixed now.

PR #203, opened 2017-03-08
restrict non-admin users to certain netblock sizes

adrianreber:
Backport from mirrormanager1:
https://pagure.io/mirrormanager/blob/master/f/server/mirrormanager/controllers.py#_655

The original mirrormanager implementation restricted IPv4 netblock sizes
to /16 and IPv6 netblock sizes to /32. This is used to prevent mirror
admins to make the whole internet their local netblock.

This change reintroduces the above code from the original mirrormanager
implementation.


adrianreber:
Fixes #71

pypingou:
Looks fine to me :)

Issue #206, opened 2017-03-29
Tracker for fullfilelist-powered UMDL


puiterwijk:
@adrianreber was working on an improved UMDL that uses fullfilelist.
This ticket is just to track that, so interested people can follow it.

adrianreber:
Running final tests for the fullfiletimelist-* support.

The goal of the fullfiletimelist-* based update-master-directory-listing (umdl) is to be able to 'scan' the master mirror without stat(). The main reason is that the filesystem trees umdl is working have become huge and to scan all those files produces lot's of IOPS just for detecting that nothing changed.

With the help of https://pagure.io/quick-fedora-mirror/ new files (fullfiletimelist-*) are created for each mirror category.

This new umdl approach relies on the correctness of those files. This means that if those files are wrong, MirrorManager will have the wrong state. But as it removes (almost) the necessity to do any stat() it it much faster. It is between 5 and 10 times faster.

The current implementation still needs to stat the directories to see if they are readable but there is a feature request (https://pagure.io/quick-fedora-mirror/issue/40) to include the readable information in the fullfiletimelist-* files.

To calculate the repomd.xml checksums and to read the -CHECKSUM files of the ISOs there is still disk access necessary (could be also via https or rsync) but only if the directory ctime actually changed (according to fullfiletimelist-*).

The fullfiletimelist-* files are detected automatically and if they do not exist umdl falls back to directory walking.

adrianreber:
This is now running on the Fedora production systems.

PR #207, opened 2017-03-31
 umdl: add fullfiletimelist-* based master scanning

adrianreber:
Disk based scanning all modules of the Fedora master mirror takes hours.
A full scan takes (at least) more than 6 hours and stat()s every
directory in the whole tree. This creates a massive amount of IOPS which
are basically only used to detect that most directories have not
changed.

With the newly created fullfiletimelist-* files for fedora-quick-mirror
(https://pagure.io/quick-fedora-mirror/) this can be changed and a nearly
stat()-less master mirror scan is possible and implemented by this commit.

Important to remember is that the result will always only be as good as
the input files (fullfiletimelist-*).

With this commit umdl detects if a 'fullfiletimelist-*' file exists and
instead of walking the directory tree only the found file is parsed. To
switch back to disk based scanning a new switch is introduced:

  --skip-fullfiletimelist
                        Do not look for a fullfiletimelist-*; actually scan
                        the filesystem

Using the fullfiletimelist-* files umdl is much faster:

 * Fedora Linux
   * disk based scanning: 48m43.167s
   * fullfiletimelist-fedora: 2m5.367s
 * Fedora EPEL
   * disk based scanning: 2m59.480s
   * fullfiletimelist-epel: 0m5.432s
 * Fedora Secondary Arches
   * disk based scanning: 80m13.825s
   * fullfiletimelist-fedora-secondary: 2m51.272s
 * Fedora Archive
   * disk based scanning: 182m49.157s
   * fullfiletimelist-archive: 2m50.970s
 * Fedora Other
   * disk based scanning: 173m24.203s
   * fullfiletimelist-alt: 10m47.376s

So it is much faster but heavily depends on the correctness of the
fullfiletimelist-* files. Currently a single stat() for every changed
directory is necessary to detect if the directory is readable or not
(pre-bitflip scenario mainly). The feature request for
quick-fedora-mirror to include this information already exists:
https://pagure.io/quick-fedora-mirror/issue/40

To read the *-CHECKSUM files for the information about the ISOs disk
access is still required (could also be downloaded via https or rsync)
and also for the checksums of the repomd.xml file for the metalink.

One point which made this whole thing more complicated than necessary is
the 'Fedora Linux' category. All other category topdir point to
/srv/pub/<module> . Only 'Fedora Linux' points to
/srv/pub/<module>/linux -> /srv/pub/fedora/linux .

The fullfiletimelist-fedora, however, starts at /fedora so that the paths
of this single category cannot be joined as easily as for the other
category. This makes the code at some places unnecessarily complicated.

### DIFF
--- a/utility/mm2_update-master-directory-list
+++ b/utility/mm2_update-master-directory-list
@@ -52,70 +52,141 @@ class MasterFilter(logging.Filter):
         record.category = cname
         return True
 
-def make_file_details_from_checksums(session, diskpath, relativeDName, D):
+
+def make_file_details_from_checksums(session, config, D, files=None):
+    """
+    This function looks for checksum files in the given directory. From those
+    checksum files the checksums are extracted and then written to
+    MirrorManager's database. With the checksums in the database
+    MirrorManager can create metalinks for those files, ISOs for example.
+
+    If the variable :files: is None this function will actually access the
+    file-system to detect if the checksum files exits. If :files: is not None
+    the function will use files to detect if checksum files exist.
+
+    :param session: SQLAlchemy session
+    :param config: umdl config
+    :param D: MirrorManager database directory object
+    :param files: optional hash with the files in the directory; if this is
+                  None, the function will actually scan the directory :D:
+    """
+
+    def _handle_checksum_line(line, checksumlen):
+        """
+        This function tries to extract the filename and the its checksum
+        from given :line:.
+        """
+
+        line = line.strip()
+        try:
+            # old style fileformat
+            # 25ceb179396f45586e373cee04bea0c436481a5b340bc24073f8178546b1e51b *Fedora-i386-20-20131211.1-sda.qcow2
+            checksum, filename = line.split()
+            algo, delimiter = None, None
+        except ValueError:
+            # new style - bsd style checksum
+            # SHA256 (Fedora-Cloud-Base-25-1.3.x86_64.qcow2) = 4942be57a97c7e6f9f420ba29555776c79ed49c603d7cabb0394bb1a790eb336
+            algo, filename, delimiter, checksum = line.split()
+
+        if algo is None:
+            # old style fileformat
+            if len(checksum) != checksumlen:
+                return None, None
+            # strip off extraneous starting '*' char from name
+            filename = filename.lstrip('*')
+            return filename, checksum
+
+        # new style - bsd style checksum
+        if (checksumlen == 32) and (algo != 'MD5'):
+            return None, None
+        if (checksumlen == 40) and (algo != 'SHA1'):
+            return None, None
+        if (checksumlen == 64) and (algo != 'SHA256'):
+            return None, None
+        if (checksumlen == 128) and (algo != 'SHA512'):
+            return None, None
+
+        filename = filename[1:-1]
+
+        return filename, checksum
+
     def _parse_checksum_file(relativeDName, checksumlen):
         r = {}
         try:
-            f = open(os.path.join(diskpath, relativeDName),  'r')
-            for line in f:
-                line = line.strip()
-                s = line.split()
-                if len(s) < 2:
-                    continue
-                if len(s[0]) != checksumlen:
-                    continue
-                # strip off extraneous starting '*' char from name
-                s[1] = s[1].strip('*')
-                r[s[1]] = s[0]
-            f.close()
+            with open(os.path.join(config.get('UMDL_PREFIX', ''),
+                relativeDName),  'r') as f:
+                for line in f:
+                    filename, checksum = _handle_checksum_line(line, checksumlen)
+                    if filename != None:
+                        r[filename] = checksum
         except:
             pass
         return r
 
     def _checksums_from_globs(relativeDName, globs, checksumlen):
+        """
+        Finds checksum files on disk (or from the filelist) and extracts
+        the checksums and places the result in a dict[filename] = checksum.
+        Finding the checksum files does a os.listdir() if the variable
+        files is None.
+
+        :param relativeDName: path to the directory (without category topdir)
+        :param globs: possible patterns containing checksums
+        :param checksumlen: number of characters the checksum has
+        :returns: dict[filename] = checksum
+        """
+
         d = {}
         checksum_files = []
         for g in globs:
-            checksum_files.extend(
-                glob.glob(os.path.join(diskpath, relativeDName, g)))
+            if files != None:
+                for f in files.keys():
+                    if re.compile(g).match(f):
+                        checksum_files.append(os.path.join(relativeDName,f))
+            else:
+                for f in os.listdir(os.path.join(config.get('UMDL_PREFIX', ''), relativeDName)):
+                    if re.compile(g).match(f):
+                        checksum_files.append(os.path.join(relativeDName,f))
         for f in checksum_files:
             d.update(_parse_checksum_file(f, checksumlen))
         return d
 
-    if diskpath is None:
-        return
+    sha1_globs = ['.*\.sha1sum', 'SHA1SUM', 'sha1sum.txt']
+    md5_globs = ['.*\.md5sum', 'MD5SUM', 'md5sum.txt']
+    sha256_globs = ['.*-CHECKSUM', 'sha256sum.txt']
+    sha512_globs = ['.*\.sha512sum', 'SHA512SUM', 'sha512sum.txt']
+    md5dict = _checksums_from_globs(D.name, md5_globs, 32)
+    sha1dict = _checksums_from_globs(D.name, sha1_globs, 40)
+    sha256dict = _checksums_from_globs(D.name, sha256_globs, 64)
+    sha512dict = _checksums_from_globs(D.name, sha512_globs, 128)
 
-    sha1_globs = ['*.sha1sum', 'SHA1SUM', 'sha1sum.txt']
-    md5_globs = ['*.md5sum', 'MD5SUM', 'md5sum.txt']
-    sha256_globs = ['*-CHECKSUM', 'sha256sum.txt']
-    sha512_globs = ['*.sha512sum', 'SHA512SUM', 'sha512sum.txt']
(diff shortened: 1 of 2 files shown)


adrianreber:
#206

adrianreber:
I know this is difficult to review, but it runs successfully in staging and I would like to get it deployed in prod. So hoping to get a review to get merged soon.

utility/mm2_update-master-directory-list
```diff
@@ -52,21 +53,32 @@ class MasterFilter(logging.Filter):
...
-def make_file_details_from_checksums(session, diskpath, relativeDName, D):
+def make_file_details_from_checksums(session, config, D, files=None):
     def _parse_checksum_file(relativeDName, checksumlen):
         r = {}
         try:
-            f = open(os.path.join(diskpath, relativeDName),  'r')
+            f = open(os.path.join(config.get('UMDL_PREFIX', ''), relativeDName),  'r')
             for line in f:
                 line = line.strip()
                 s = line.split()
```
jeremycline: This whole block might be easier to read if you did some tuple unpacking here, something like:
```
try:
     field1, field2 = line.split()
     field3, field4 = None, None
except ValueError:
     # Needs more than 2 values to unpack the tuple
    field1, field2, field3, field4 = line.split()  # If this fails it's an unexpected format that results in a ValueError
```
Obviously with more enlightening field names, but you get the idea. It's up to you.

utility/mm2_update-master-directory-list
```diff
@@ -76,46 +88,54 @@ def make_file_details_from_checksums(session, diskpath, relativeDName, D):
         d = {}
         checksum_files = []
         for g in globs:
-            checksum_files.extend(
-                glob.glob(os.path.join(diskpath, relativeDName, g)))
+            if files != None:
```
jeremycline: I believe you _could_ rewrite this so you don't need to duplicate code
```
file_paths = files.keys() if files else os.listdir(os.path.join(config.get('UMDL_PREFIX', ''), relativeDName))
for f in file_paths:
    if re.compile(g).match(f):
          checksum_files.append(os.path.join(relativeDName,f))
```
Total nick-pick though.

utility/mm2_update-master-directory-list
```diff
@@ -76,46 +88,54 @@ def make_file_details_from_checksums(session, diskpath, relativeDName, D):
         d = {}
```
jeremycline: I recommend documenting the arguments and return value of this function

utility/mm2_update-master-directory-list
```diff
@@ -52,21 +53,32 @@ class MasterFilter(logging.Filter):
         record.category = cname
         return True
 
-def make_file_details_from_checksums(session, diskpath, relativeDName, D):
+def make_file_details_from_checksums(session, config, D, files=None):
```
jeremycline: I recommend documenting the arguments and return value of this function

utility/mm2_update-master-directory-list
```diff
@@ -138,6 +158,7 @@ def make_file_details_from_checksums(session, diskpath, relativeDName, D):
                 size=size)
             session.add(fd)
             session.commit()
+            logger.debug("Added checksum for %s to database" % f)
```
jeremycline: I _believe_ if you pass ``f`` as an argument rather than substituting it - ``logger.debug("Added checksum for %s to database", f)`` - it will cause the substitution to occur when logging at the debug level is on.

That's probably not going to be a huge performance gain in the grand scheme of things, but it might add up. Just an FYI.

utility/mm2_update-master-directory-list
```diff
@@ -423,6 +455,84 @@ def short_filelist(files):
         return files
 
 
+def sync_category_directory(
```
jeremycline: I recommend documenting this function's purpose, parameters, and return value

utility/mm2_update-master-directory-list
```diff
@@ -423,6 +455,84 @@ def short_filelist(files):
...
+
+        # stating changed files to see if they are readable
+        s = None
+        try:
+            s = os.stat(os.path.join(
+                config.get('UMDL_PREFIX', ''),
+                topdir,
+                relativeDName))
+        except OSError:
+            logger.debug("Hmm, stat()ing %s failed. Theoretically this cannot happen." % relativeDName)
```
jeremycline: Maybe log this at a warning or error level? It seems like an exceptional, entirely unexpected result.

utility/mm2_update-master-directory-list
```diff
@@ -423,6 +455,84 @@ def short_filelist(files):
...
+
+        if s:
+            mode = s.st_mode
+            readable = not not (mode & stat.S_IRWXO & (stat.S_IROTH|stat.S_IXOTH))
```
jeremycline: This definitely could use a comment. I think what it does is check to see whether or not the file is readable or executable by others?

Applying ``not`` twice is also very odd. Maybe you just want to cast it to a boolean with ``bool``?

utility/mm2_update-master-directory-list
```diff
@@ -423,6 +455,84 @@ def short_filelist(files):
...
+            if D.readable != readable:
```
jeremycline: Since this block executes if they aren't equal and its result is to make them equal, I think it's okay to drop the ``if`` entirely, right?

utility/mm2_update-master-directory-list
```diff
@@ -634,8 +743,111 @@ def sync_directories_from_disk(
     sync_category_directories(
         session, config, diskpath, category, category_directories)
 
+    if category.delete_directories:
+        nuke_gone_directories(session, diskpath, category)
+
+
+def sync_directories_from_fullfiletimelist(
```
jeremycline: I recommend documenting the purpose, arguments, and return value(s) of this function

utility/mm2_update-master-directory-list
```diff
@@ -634,8 +743,111 @@ def sync_directories_from_disk(
...
+
+    if len(filelist) < 1:
+        # not a single element found in the glob
+        return False
+
+    # blindly take the first file found by the glob
+    logger.info('Get fullfiletimelist %s' % filelist[0])
+    # A hash with directories as key and ctime as value
+    ctimes = {}
+    # As hash with directories as key and
```
jeremycline: s/As/A/ I think

utility/mm2_update-master-directory-list
```diff
@@ -634,8 +743,111 @@ def sync_directories_from_disk(
...
+            row = m.readline()
+            # only rows with at least 4 columns are what we are looking for
+            # 'ctime\ttype\tsize\tname'
+            if len(col) >= 4:
+                # only files (type: f)
+                if col[1] == 'f':
+                    # put all the information in a dict with path as key
+                    tmp = os.path.dirname(_handle_fedora_linux_category(col[3]))
+                    tmp = os.path.join(diskpath, tmp).replace(umdl_prefix, '')
+                    tmp = unicode(tmp)
```
jeremycline: It'd be good to be explicit about the encoding you're expecting here

utility/mm2_update-master-directory-list
```diff
@@ -755,12 +970,29 @@ def main():
...
-            sync_directories_from_disk(session, config, i['path'], category)
+            found = False
+            if not options.skip_fullfiletimelist:
+                if category.name == 'Fedora Linux':
+                    # This seems strange... and it really is strange
+                    # The 'Fedora Linux' category is the only category
+                    # where the topdir is wrong. This removes '/linux'
+                    # at the end to fix it for fullfiletimelist walking
+                    path = i['path'][:-6]
+                    path = i['path']
```
jeremycline: Did you mean to overwrite your sliced path here? This doesn't look right to me.

dustymabe:
anyone else able to review this?

utility/mm2_update-master-directory-list
```diff
@@ -52,21 +53,32 @@ class MasterFilter(logging.Filter):
         record.category = cname
         return True
 
-def make_file_details_from_checksums(session, diskpath, relativeDName, D):
+def make_file_details_from_checksums(session, config, D, files=None):
     def _parse_checksum_file(relativeDName, checksumlen):
         r = {}
         try:
-            f = open(os.path.join(diskpath, relativeDName),  'r')
+            f = open(os.path.join(config.get('UMDL_PREFIX', ''), relativeDName),  'r')
```
adrianreber: Thanks for all the reviews, could you give me an example what you mean to use a context manager. Would

    with open(os.path.join(config.get('UMDL_PREFIX', ''), relativeDName),  'r') as f:

be enough? Then the close() could also be removed?

jeremycline: Yup, that will automatically close the file when exiting the block of code (even if an exception occurs)

utility/mm2_update-master-directory-list
adrianreber: If I remember it correctly I was looking for a way to avoid duplicate code when writing this. But I usually try to avoid long python lines like you proposed as I find it harder to read in this case than the duplicate code.

utility/mm2_update-master-directory-list
adrianreber: Yes a comment would be good. Unfortunately I am not totally sure what that does. That is the same code as in the original MirrorManager code base:

https://pagure.io/mirrormanager/blob/master/f/server/update-master-directory-list#_494

I will leave it as it is. If the fullfiletimelist support for global unreadable is ready I will change this as the stat() will no longer be required. For now I will leave it as it is.

https://pagure.io/quick-fedora-mirror/issue/40

utility/mm2_update-master-directory-list
adrianreber: Thanks for noticing. That was some old test code which needs to be removed.

adrianreber:
Thanks for all the reviews. I tried to address all of the comments I was able to.

I updated the PR and rerun the tests in staging.

adrianreber:
Any other reviews, comments? If not, I would like to merge and make a new release.

jeremycline:
I think at this point it's fine to merge it 👍

PR #214, opened 2017-08-04
Revert namespace change - this breaks namespace aware xml parsers

puiterwijk:
This reverts a small part of dad4e76899b28755c39d4f494f618a5a4037ebf7.
Changing the XML namespace needs to be more coordinated.

Signed-off-by: Patrick Uiterwijk <patrick@puiterwijk.org>


adrianreber:
I guess there were problems when this was deployed in production. Changing it back seems correct. 👍

puiterwijk:
@adrianreber yes there were :). Yum suddenly started shouting it was missing the timestamp, which is in the mirrormanager namespace :). Thanks!

PR #218, opened 2017-09-10
Correctly detect repositories

adrianreber:


utility/mm2_update-master-directory-list
```diff
@@ -290,19 +290,6 @@ def make_repo_file_details(session, diskpath, relativeDName, D, category, target
         session.commit()
 
 
-def move_repository_from_development(prefix, arch, newdir):
```
pypingou: was this not called anywhere?

pypingou:
One question otherwise looks fine to me :)

utility/mm2_update-master-directory-list
adrianreber: This function has never been used. Even going back to the version 0.0.1 tag it is not used anywhere.

This step is always done manually after a release.

pypingou: Alright :)

PR #219, opened 2017-09-10
umdl: only create repositories for 'Everything'

adrianreber:
This is a fix for

 https://pagure.io/fedora-infrastructure/issue/6286

This hardcodes a list ('Cloud', 'Server', 'Workstation') to skip before
creating a repository. This should leave 'Everything' and other strings,
like used in the updates directory structure.

Not totally sure of the side effects of this change.


adrianreber:
This includes the commits from PR #218

PR #220, opened 2017-09-20
detect and setup mirrorlist/metalinks for modular Fedora

ausil:
in the new Modular world we need to setup a new structure for
delivering content. it will look something like bellow

/pub/fedora/linux/modular/
/pub/fedora/linux/modular/development
/pub/fedora/linux/modular/development/bikeshed
/pub/fedora/linux/modular/development/27
/pub/fedora/linux/modular/releases
/pub/fedora/linux/modular/releases/test
/pub/fedora/linux/modular/updates
/pub/fedora/linux/modular/updates/27
/pub/fedora/linux/modular/updates/testing
/pub/fedora/linux/modular/updates/testing/27

under each releasever 27, 28 bikeshed etc will be a directory for
the consumable product Server, Workstation, Atomic, etc. For each
consumable product we will have to setup a new set of prefixes. We
are starting with Server.

Signed-off-by: Dennis Gilmore <dennis@ausil.us>

### DIFF
--- a/mirrormanager2/lib/repomap.py
+++ b/mirrormanager2/lib/repomap.py
@@ -27,6 +27,7 @@ def repo_prefix(path, category, ver):
     # assign shortnames to repositories like yum default mirrorlists expects
     isDebug = u'debug' in path
     isRawhide = u'rawhide' in path
+    isBikeshed = u'bikeshed' in path
     isDevelopment = is_development(path) is not None
     isSource = u'source' in path or u'SRPMS' in path
     isUpdatesTesting = u'updates/testing' in path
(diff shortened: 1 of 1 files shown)


adrianreber:
Looks good. It is always difficult to test MM changes so if it is totally correct we will start to see once the actual files are available.

ausil:
Thanks, this is a blocker to delivering Modular server for Fedora 27

adrianreber:
I can make a new release and install it after the beta freeze. Or is that already too late?

ausil:
Getting it in as a freeze exception would be awesome, that way we can get content in and tested, along with fixing any snags we hit.

adrianreber:
As soon as files start to appear in the modular directories I can test it in staging.

Issue #210, opened 2017-05-03
Filtering arches from repo creation


nirik:
This is a upstream tracker for 
https://pagure.io/fedora-infrastructure/issue/5186

basically we need a way to tell mirrormanager that some arches are available only in archives and shouldn't be listed for anything newer when making repositories. 

See downstream issue for more info...

adrianreber:
Kind of fixed. With PR #223 the arches are filtered out from the overview if marked as 'publiclist=false' in the database. The arches are not actually filtered out from the repository creation but removing certain arches from there was what actually was requested in the original ticket.

PR #223, opened 2017-10-18
publiclist: hide disabled arches and products

adrianreber:
The original MirrorManager code base had the feature to hide certain
arches and products from the publiclist. The database fields still exist
and are even set to 'false' for certain arches.

There is the ticket https://pagure.io/fedora-infrastructure/issue/5186
which tried to remove unused architectures from the publiclist
(sparc and sparc64). Using the now unused database column 'publiclist'
seems like the right thing to do to hide certain arches (including
source) from the publiclist. RPM Fusion which still uses the original
MirrorManager code base is a good example how it used to look:

 http://mirrors.rpmfusion.org/mm/publiclist/

The architecture 'source' is not displayed, which has in both instances
(Fedora and RPM Fusion) publiclist = False. It still can be seen in
Fedora:

 https://admin.fedoraproject.org/mirrormanager/

I tested my changes in Fedora's staging instance and marked sparc,
sparc64 and ia64 as publiclist=false in the database:

 https://admin.stg.fedoraproject.org/mirrormanager/

### DIFF
--- a/mirrormanager2/app.py
+++ b/mirrormanager2/app.py
@@ -192,8 +192,10 @@ def inject_variables():
 def index():
     """ Displays the index page.
     """
-    products = mmlib.get_products(SESSION)
-    arches = mmlib.get_arches(SESSION)
+    # publiclist=True filters out all results which have
+    # publiclist set to False
+    products = mmlib.get_products(SESSION, publiclist=True)
+    arches = mmlib.get_arches(SESSION, publiclist=True)
     arches_name = [arch.name for arch in arches]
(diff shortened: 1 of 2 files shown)


adrianreber:
Fixes #210

mdomsch:
I love that my old features are still being found like easter eggs!

adrianreber:
@mdomsch, thanks for all the easter eggs ;-) . It really helps to have RPM Fusion still running on the original code base to see how some things were supposed to work.

PR #224, opened 2017-10-19
repomap: fix modular updates repository

adrianreber:
To make umdl detect the modular updates repositories the repomap prefix
detection had to be moved under isModular. Now some update repositories
are under isUpdates* and some are under isModular.

repomap.py has become unreadable and should actually be a
configuration file, somehow.


adrianreber:
The change looks more complicated than it is as I had to change the indentation level as some ifs have moved around. So basically only changes to some ifs.

adrianreber:
@ralphbean

adrianreber:
Any reviewers? This already verified in Fedora's production environment,

pypingou:
👍 for me

PR #228, opened 2017-11-22
Fix all tests

adrianreber:
Most of the tests seemed to be broken for quite some time already. This PR fixes all included tests to work again and reduces the size of the testdata directory from 37M to 1.4M.


pypingou:
Looks alright to me but I didn't go into all the details tbh

adrianreber:
Probably hard to review, but what would be necessary to run those tests after each commit?

PR #134, opened 2015-09-28
Add the possibility to kill rsync after some time

pypingou:

Issue #77, opened 2015-05-16
Introduce timeout check during run_rsync()


adrianreber:
Currently the crawler timeout is ignored if crawling via rsync. The possibility to specify a rsync timeout value on the command-line exists and is used but the rsync timeout options has another meaning:

```
 --timeout=TIMEOUT
              This option allows you to set a maximum I/O timeout in seconds.
              If no data is transferred for the specified time then rsync will exit.
              The default is 0, which means no timeout.
```

The crawling, however, should stop with an error if it did not finish during the specified timeout like it does with HTTP or FTP.

See #53 and #55

pypingou:
This sound familiar as well, didn't you work on this? (or was it only for non-rsync method?)

adrianreber:
It was only for non-rsync methods. Right now the rsync binary is spawned but we have no control how long it will run. I have not really looked into it how a timeout could be controlled. Starting a thread for it sounds a bit like overkill but I do not know if there is another option to control the duration of a spawned binary.

PR #229, opened 2017-11-23
crawler: use timeout also on rsync crawls

adrianreber:
Over two years ago PR #134 added support to abort the rsync based crawls if it takes too long. This commit now uses this existing timeout support to the crawler. The timeout is set to 90% of the remaining crawl time for the current host. This is also related to #77.

### DIFF
--- a/mirrormanager2/lib/sync.py
+++ b/mirrormanager2/lib/sync.py
@@ -29,6 +29,7 @@ import tempfile
 import errno
 import threading
 
+
 def check_timeout(logger, p, timeout, e):
     ''' Check if the process is running and kill it if so. '''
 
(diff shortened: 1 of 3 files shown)


mirrormanager2/lib/sync.py
```diff
@@ -59,21 +90,28 @@ def run_rsync(rsyncpath, extra_rsync_args=None, logger=None, timeout=None):
         stderr=devnull,
         close_fds=True,
         bufsize=-1
+
```
pypingou: intended?

pypingou:
Looks alright to me :)

mirrormanager2/lib/sync.py
adrianreber: no, thanks for spotting

PR #233, opened 2017-11-24
tests: add possibility to skip network

adrianreber:
After adding the reactivated test cases to the spec file a few network
dependent (id.fedoraproject.org) test cases started to fail. This skips
those tests if MM2_SKIP_NETWORK_TESTS=1.


pypingou:
👍 for me

PR #235, opened 2018-01-30
tests: fix broken test results

adrianreber:
The container setup to run the tests on centos-7 fedora-rawhide and
fedora-latest was using yum to install the packages on all systems as
that made it possible to install all three test scenarios with one
command. This no longer works as rich dependencies are not resolved by
yum and thus the test environment could no longer be correctly set up.

This commit introduces a shell script to set up the container based on
the ${DISTRO}.

Another change is to decrease the timeout for the timeout run_rsync test
case.


adrianreber:
Thanks.

Issue #239, opened 2018-02-23
check_for_base_dir only supports http - make it also work with https urls


adrianreber:

adrianreber:
Fixed in #240

PR #240, opened 2018-02-24
crawler: support https only hosts

adrianreber:
Until now mirrormanager supported HTTPS only sites but implicitly
expected that the same URL as the HTTPS URL is also available via HTTP.
And although the mirror is using a HTTPS URL mirrormanager actually used
HTTP to scan the mirror.

This change adds HTTPS support at a few places so that HTTPS only
mirrors are also supported.


adrianreber:
Fix for #239

adrianreber:
Thanks, that was fast.

Issue #238, opened 2018-02-22
Please use "site", "host" and "mirror" consistently


jasontibbitts:
I received a complaint that quick-fedora-mirror wasn't clear in its documentation of the variables regarding doing mirrormanager checkins, so I had a look at mirrormanager to see exactly what it calls various things.  Unfortunately it's not consistent.

When you look at the "My Sites" page, you get a list of "sites" and grouped together the list of "mirrors" beneath those.

When you click on one of those sites, you get a form with "Site name", "Site Password", etc. and a list of "hosts" (not mirrors) below.

If you click on a host/mirror from either of those pages, you get to a form which has "Site name" again, but this isn't the same "Site name" from the site page.

So changing to "Host name" on the host page would make sense, as well as using the word "host" instead of "mirror" on the "My Sites" page (though that's more minor).

adrianreber:
Fixed in #241

PR #241, opened 2018-02-24
Use "site", "host" and "mirror" consistently

adrianreber:
The terms "site" and "host" have a special meaning in MirrorManager. One
'Site' can consist of multiple 'Host's but the terms were not used
consistently. This tries to address that by renaming 'Site name' in the
host to 'Host name' and by renaming 'Mirror' to 'Hosts' when showing the
overview.


adrianreber:
@jasontibbitts  Fix for #238

jasontibbitts:
Hey, thanks.  I was going to try to get around to giving a patch eventually.  I guess those are the only two changes really needed, but I didn't go through the entire interface.

pypingou:
:+1: for these two

PR #243, opened 2018-03-14
repomap: more modular repository detection logic

adrianreber:
The whole repomap logic is much too complicated and this might have
already existing in some previous versions of repomap.py.

This change has already been verified in Fedora's production
MirrorManager instance.


adrianreber:
@nirik This is the fix to correctly detect the ```fedora-modular-release``` repositories.

nirik:
+1. thanks for fixing this!

PR #244, opened 2018-03-14
crawler: Correctly calculate the remaining time

adrianreber:
The calculation for the rsync timeout was completely wrong. Instead of
calculating the remaining time for the current crawler thread the time
the thread was running was calculated. Which was most of the time 1
second for the first crawled category. This lead to rsync being aborted
after 1 second and the crawler switched to http(s).

With this the remaining time of the crawler thread is now calculated
correctly.

### DIFF
--- a/utility/mm2_crawler
+++ b/utility/mm2_crawler
@@ -891,19 +891,19 @@ def try_per_category(
     timeout_check()
 
     # Get remaining time this crawl thread has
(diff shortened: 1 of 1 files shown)


adrianreber:
This is the reason I had a closer look at the timeout: https://pagure.io/fedora-infrastructure/issue/6778

Issue #249, opened 2018-05-13
Category based scanning empties all other categories on host


adrianreber:
With the recent switch in Fedora's infrastructure to category based crawling a crawler bug was uncovered which removes all directories (HCD) from a host which are not part of the current category being scanned. The function sync_hcds() should only work with the specified categories instead of all categories.

adrianreber:
Closing as the fix was merged

PR #250, opened 2018-05-13
crawler: only update directories of the current category

adrianreber:
In one last place (hopefully) the crawler was not yet category based
crawling aware. Instead of removing the not up to date directories from
the current category it removed the directories from all existing host
categories. The means that whenever one category is scanned, all other
categories are automatically empty.

Instead of looking at all categories of a host the crawler now only
looks at those categories specified on the command-line.


adrianreber:
Fix for #249

PR #253, opened 2018-09-04
Migrate to new geoip API

### LINKED ISSUE #248: GeoIP legacy is EOL
https://support.maxmind.com/geolite-legacy-discontinuation-notice/

Our GeoLite Legacy databases are being discontinued
 

We have decided to discontinue our GeoLite Legacy databases effective January 2, 2019. We will stop updating the databases before that date (see “General Information” below), and the databases will no longer be available for download on January 2, 2019. This was a difficult decision for us and we have created this page to provide more information and answer any immediate questions you may have.

===============
It looks like they are freezing updates to the geolite-legacy so we will need to use the geolite2 or geoip2 library calls.

adrianreber:
Resolves #248


adrianreber:
Any reviewers?

utility/mm2_generate-worldmap
```diff
@@ -80,6 +80,7 @@ def lookup_host_locations(config):
 
 def doit(options, config):
 
+    gi = geoip2.database.Reader('/usr/share/GeoIP/GeoLite2-City.mmdb')
```
pypingou: Wondering if that will always be in the same location

pypingou:
One question but otherwise not much standing out

utility/mm2_generate-worldmap
adrianreber: Right. I was wondering the same. But as we used to hard-code the location to the old location I thought that we just keep doing it. If we ever need to change it, it will be minimal effort to extend to be configurable or command-line parameter. (and I am pressing the wrong buttons :-)

adrianreber:
thanks

Issue #251, opened 2018-07-05
yum 3 will be deprecated in Fedora 29


adrianreber:
According to https://pagure.io/releng/issue/7588 yum 3 will be deprecated in Fedora 29.

Mirrormanager currently does at least in four files:

```import yum.repoMDObject```

Need to figure out if this still works with DNF.

Conan-Kudo:
This does not work in DNF. There is `python-pyrpmmd` which provides that exact API for Python 2 and Python 3.

Alternatively, you can use librepo's Python bindings.

adrianreber:
@Conan-Kudo thanks for the pointers, good to know

PR #254, opened 2018-09-11
rpmmd: switch from yum.repoMDObject pyrpmmd

adrianreber:
With the coming deprecation of yum in Fedora this changes MirrorManager
to use rpmmd instead of yum to read the repomd.xml files.

Fixes: #251


adrianreber:
@bowlofeggs : Thanks!

PR #256, opened 2018-10-24
Use InputRequired() instead of Required()

adrianreber:
The documentation of wtforms says that Required() is deprecated and that
either DataRequired or InputRequired() should be used for the
validators. This switches all uses of Required() to InputRequired()
which also fixes two broken test cases.


adrianreber:
Trying to work on something else I discovered that the tests are broken on F29, this is an attempt to fix the broken errors.

adrianreber:
Thanks!

Issue #252, opened 2018-07-22
Invalidate mirror when toggling private flag


adrianreber:
There have been occasions that a mirror had the private flag toggled which resulted in a bad mirror state. Especially if a mirror is switched from private mirror (only updated via report_mirror) to a public mirror. Until the mirror is crawled for the first time it might be in a unknown state and therefore it might be nice to invalidate a mirror until the first crawl to make sure the mirror is set up correctly.

PR #257, opened 2018-10-28
Toggle private

adrianreber:
Implementation for #252


tests/test_ui_app.py
```diff
@@ -1683,6 +1683,159 @@ def test_auth_logout(self):
             self.assertTrue(
                 '<h2>Fedora Public Active Mirrors</h2>' in data)
 
+    def test_toggle_private_flag(self):
```
Zlopez: I recommend to split these tests, so it is more clear, what are you trying to test.

adrianreber: Thanks for the review. I split the one test into two different tests.

PR #260, opened 2019-01-05
Enable MirrorManager2 to be built using Python 3 for Fedora

Conan-Kudo:
This changes all the shebangs to an unversioned one that works in
venvs for development, while adjusting the spec file to change all
the shebangs to the correct Python interpreter during the package
build.

Tests are also now run using Python 3 on Fedora, and Python 2 on CentOS 7.

This strategy is similar to what is used for Pagure.

Signed-off-by: Neal Gompa <ngompa13@gmail.com>

PR #261, opened 2019-01-10
Fix tests with python3

adrianreber:
This PR fixes the tests which were no longer working in #260.


Conan-Kudo:
@adrianreber Looks good to me, could we make a release of MirrorManager2 after merging this so that it can be switched to Python 3 in Fedora (and potentially in prod, if that's running on Fedora)?

adrianreber:
@Conan-Kudo I will make a release soon. There is one more PR I want to merge.

PR #262, opened 2019-01-21
Use Tox and Jenkins

abompard:
You need to have the `python{2,3}-pyrpmmd` and `python{2,3}-fedora-flask` rpms installed.

### DIFF
--- a/.cico.pipeline
+++ b/.cico.pipeline
@@ -1,0 +1,1 @@
+fedoraInfraTox { }
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,21 @@
+[run]
+# Track what conditional branches are covered.
+branch = True
+#include =
+#    mirrormanager2/*
+
+[report]
+# Fail if the coverage is not 100%
+fail_under = 42
+# Display results with up 1/100th of a percent accuracy.
+precision = 2
+exclude_lines =
+    pragma: no cover
+
+    # Don't complain if tests don't hit defensive assertion code
+    raise AssertionError
+    raise NotImplementedError
+
+    if __name__ == .__main__.:
+omit =
+    mirrormanager2/tests/*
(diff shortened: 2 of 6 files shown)


adrianreber:
Thanks for the PR, but I do not know what this is about. I know what jenkins is but how is this connected to MirrorManager and what does tox do? And why? 😀

abompard:
Hey Adrian, I realize this is rather incomplete as a description, but I've done too many of those today ;-)

With Brian Stinson and the rest of the team, we're setting up CentOS CI (based on Jenkins) to check all our applications in Fedora Infra.
[Tox](http://tox.readthedocs.io/) is a very popular test runner wrapper for python that creates virtualenvs for differerent versions of python (and possibly dependency versions) and runs any test command in these environment (be it nose, pytest, plain unittest, flake8, etc). Since it's so popular most of our apps already use it, and since CentOS CI would like to have as few entry points as possible to run the tests in our apps, we decided that we'd just setup Tox for all our Python apps and teach CentOS CI to run tox.

At the moment mirrormanager isn't ported on Python3 (unless I'm mistaken), so tox isn't as useful as can be yet, but this PR also sets up coverage reports and diff-cover (which checks if a PR is increasing or decreasing coverage). It also allows building the docs from within Tox (with `tox -e docs`), so you'll know if a PR breaks the docs.

Conan-Kudo:
@abompard MirrorManager2 is now a Python 3 application since 0.9.0 on Fedora.

abompard:
OK, I can't get the tests to pass because of a dependency on Flask-XML-RPC which seems to be a Python2-only library. Am I missing something?

Conan-Kudo:
@abompard flask-xml-rpc has a patch in the Fedora package to make it Python 3 compatible, but it was forked in PyPI instead. In PyPI, it is `flask-xml-rpc-re` (Flask-XML-RPC-Reloaded).

ryanlerch:
@abompard going to close this one off, as it seems like we don't need it anymore (with your current changes to using tox and github actions)

Issue #264, opened 2019-03-14
Auto-disable of mirrors is broken


adrianreber:
Since running a dedicated crawler process for each category each crawl of a non existing category on a mirror reset the crawl failure counter and so we never reached the point of a mirror being auto-disabled.

PR #265, opened 2019-03-14
crawler: correctly auto-disable mirrors

adrianreber:
When using the crawler to scan each category separately a non-existing
category resets the failure counter to 0. This way almost no mirror will
ever be auto-disabled (except broken mirrors having all categories configured).


adrianreber:
Fixes #264

PR #266, opened 2019-03-22
Use protobuf instead of pickle for backend - mirrorlist data exchange

adrianreber:
This is an experiment to see if protbuf is better suited for data transfer from the backend to the mirrorlist server.


adrianreber:
Need to install additional packages for the travis runs to work with the protobuf files.

Issue #267, opened 2019-04-19
Empower globally available mirrors to be accurately represented


survient:
Issue #194 already covers providing a way for a mirror to be marked as "global" but with a low priority. One issue I see with this is that a globally available mirror may have presence in areas with an abundant number of local mirrors but due to being deemed "low-priority" it may never get any hits and winds up being purely a fail-over mirror.

Another issue I see comes up when you have a globally available mirror load balanced via DNS using a feature of a platform such as an F5 GTM load balancer. To quote my message to the fedora mirror mailing distro:

> Our DNS entries are load balanced via F5 GTM load
> balancers such that we use the same fqdn for our mirrors globally, but
> caching nameservers pull different entries depending on their
> latency/region. The problem is that we aren't able to add the same URL to
> multiple hosts as can be seen with my first issue. What I'd like to see is
> the ability for a single "Site" to have multiple hosts defined that are able
> to re-use the same URL in the case that global server load balancing via DNS
> is being used, possibly only with fedora admin approval and/or unique ASNs
> are used.

While I did suggest having the ability for multiple hosts to use the same URL, I'm not intimately familiar with the MirrorManager's inner workings so there may be a better way to go about this. I wanted to open this case at the recommendation of the fedora mirror mailing distro to brainstorm on this idea and see if a feasible solution could be crafted that has a reasonable chance of getting implemented within MirrorManager.

Alternatively possibly something along the lines of allowing the "Country" field to be overloaded, and/or the ASN field to be overloaded? Maybe some association between the two so that there is an ASN defined per country(unless MirrorManager would be intelligent enough to do this automatically)?

adrianreber:
If someone provides a patch I am pretty sure we will include this and in theory it should not be too difficult to implement.

The problem I am seeing is that the URL is marked as something unique and there are probably multiple places in MirrorManager where the current code relies on that uniqueness.

This is probably also only useful for a small number of mirrors. If somebody is interested in exploring and implementing this that would be great.

mdomsch:
The "low priority" "always_up2date" mirror configuration combination of options was intentional.  We put the Fedora Infrastructure-maintained mirrors in that configuration, as mirrors of last resort. They're at the bottom of the prioritized mirrorlist then.  They shouldn't be used directly (that's why we have all the mirrors) unless the mirrors are somehow all stale, or there are no mirrors in that country or continent.

[...]

PR #268, opened 2019-05-10
Let mirrors specify multiple countries

mdomsch:
https://github.com/fedora-infra/mirrormanager2/issues/267
requests the ability for a mirror (really, a globally load balanced
set of mirrors behind a single DNS name) to be able to specify the
countries of their actual mirrors.

This patch lets Host.country be a comma-separated list of countries.

### DIFF
--- a/mirrormanager2/lib/mirrorlist.py
+++ b/mirrormanager2/lib/mirrorlist.py
@@ -171,24 +171,27 @@ def populate_directory_cache(session):
                 category_topdir_cache[category_id]:]
             del repos
 
+        countries = []
         if country is not None:
             country = country.upper()
+            countries = country.split(u',')
 
         if not siteprivate and not hostprivate:
             add_host_to_set(cache[directoryname]['global'], hostid)
 
-            if country is not None:
+            for country in countries:
                 if country not in cache[directoryname]['byCountry']:
                     cache[directoryname]['byCountry'][country] = set()
                 add_host_to_set(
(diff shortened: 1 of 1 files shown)


ott:
I think this is a useful feature, in particular for smaller countries and for people who live in a border region. Several use cases in Europe come to mind. However, I think it would be preferred in the relational model to make the country column a table instead of a comma separated string. Only smaller changes would be required for it to add a model definition and migration the current data.

@mdomsch What do you think about this idea? And are you still interested in this pull request? If not, I would try to adopt it.

mdomsch:
I'm not directly involved in mirrormanager maintenance anymore.  The current maintainers would be in a better position to comment as to the direction they want to see here (if any).

abompard:
Hey! It looks like hosts and countries are already linked by a relationship table: [HostCountry](https://github.com/fedora-infra/mirrormanager2/blob/78656c34cb38c130fdea83ff3abc9c4c41378521/mirrormanager2/lib/model.py#L923). So, hosts can already have multiple countries. Am I missing something? (I only did maintenance on the code a few months ago, I'm not aware of all use cases).

abompard:
Ah, I see, there's a confusion between `HostCountry` and `HostCountryAllowed` in the UI. The list of allowed countries in the host form populates the `.countries` attribute, which is for `HostCountry`, not `HostCountryAllowed`. This needs to be fixed.

Issue #181, opened 2016-06-23
Add host to xmlrpc logging


adrianreber:
The logging introduced in #175 should also log which host submitted the data via report_mirror.

PR #269, opened 2019-05-11
Fix report_mirror with python3

adrianreber:
This tries to fix https://bugzilla.redhat.com/show_bug.cgi?id=1707684 which reported that report_mirror does not work. This now works with python2 and python3.

At the same time this adds some debug output to the xmlrpc checkin code. to better see which hosts are using pickle and which are using json.

### DIFF
--- a/client/report_mirror
+++ b/client/report_mirror
@@ -1,7 +1,9 @@
 #!/usr/bin/env python
 
 from __future__ import print_function
-import os, sys
+import os
+import sys
+import copy
 import json
 try:
     import configparser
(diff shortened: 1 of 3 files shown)


adrianreber:
Fixes #181

client/report_mirror
```diff
@@ -315,13 +305,10 @@ def main():
...
-        #   print "Connecting to %s" % item.config['global']['server']
+        #  print("Connecting to %s" % item.config['global']['server'])
         server = xmlrpclib.ServerProxy(item.config['global']['server'])
-        data = None
-        try:
-            data = base64.urlsafe_b64encode(bz2.compress(p))
-        except AttributeError:
-            data = urlsafe_b64encode(bz2.compress(p))
+
+        data = base64.urlsafe_b64encode(bz2.compress(p.encode())).decode()
```
adrianreber: @pypingou this is the main change to make report_mirror python3 work with a server in python2. Does that make sense or is it just luck that it works with this change?

pypingou: Depending on the datatype it can be it :)

PR #272, opened 2019-10-04
Removes fedmenu from mirrormanager2

### LINKED ISSUE #271: Remove fedmenu from mirrormanager2
The fedmenu project is not really maintained anymore and will be soon archived. We should remove fedmenu from this application.

https://github.com/fedora-infra/mirrormanager2/search?q=fedmenu&unscoped_q=fedmenu

alfredmyers:
Fixes #271 "Remove fedmenu from mirrormanager2"

From looking at the revision history, it is my understanding that fedmenu brought in a dependency to jQuery 1.11.2 with it as can be seen in 888245cc76ad5c44f969459b0747acafb3bbfc07.

I wasn't sure I could remove the reference to it since bootstrap takes a dependency on jQuery and I don't know if it takes care of bringing it in itself.

### DIFF
--- a/mirrormanager2/templates/fedora/master_noauth.html
+++ b/mirrormanager2/templates/fedora/master_noauth.html
@@ -85,18 +85,6 @@
     <script type="text/javascript" src="https://apps.fedoraproject.org/global/fedora-bootstrap-1.0.2/fedora-bootstrap.js">
     </script>
 
-
-    {% if config['FEDMENU_URL'] %}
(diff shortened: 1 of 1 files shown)


adrianreber:
Thanks for this PR. Unfortunately I am also not sure if jquery can be removed. Grepping through the source it seems not really necessary and removing it should be no problem. But then I also have no idea how all the JS works.

PR #276, opened 2019-11-11
Only query database once for mirrorlist export

adrianreber:
If the mirrorlist data exporter is writing pkl and protobuf files at the same time some database queries are running twice, which can take a lot of time. This change moves all database queries in the populate_all_caches() function which is the correct place. Previously some queries were run in populate_all_caches() and some in dump_ccaches().

This commit also includes a single line change to mm2_update-mirrorlist-server to forward all parameters to the actual data exporter script.

Verified in Fedora's mirrormanager2 instance.


adrianreber:
This was a duplicate PR.

Issue #297, opened 2021-07-12
admin-only categories


carlwgeorge:
Currently mirrormanager allows any mirror to host any category.  I would like to have the ability to have a category that only an admin can add to a mirror.

adrianreber:
That does not sound too complicated. Extending the `category` table with an `admin_only` flag should work.

adrianreber:
This can now be tested in Fedora's staging environment.

PR #298, opened 2021-07-13
Support admin only categories

adrianreber:
See #297


adrianreber:
rawhide is currently still broken due to the python 3.10 rebuild

PR #305, opened 2021-12-30
Use socket.getaddrinfo instead of socket.gethostbyname

ott:
socket.gethostbyname only returns a single IPv4 address. Any additional
IPv4 addresses and all IPv6 addresses are ignored. As a consequence,
hosts without IPv4 addresses are ignored and any additional information
from additional IPv4 addresses is ignored. If IPv4 addresses are rotated
by the implementation of socket.gethostbyname, for example, common with
DNS servers and resolvers, the geolocation information for the host can
change without every call of socket.gethostbyname.

It seems better to consider all addresses, both all IPv4 and IPv6
addresses, and to summarize the geolocation information when only a
single result is required.

### DIFF
--- a/mirrormanager2/crawler/continents.py
+++ b/mirrormanager2/crawler/continents.py
@@ -80,14 +80,48 @@ def check_continent(config, options, session, categoryUrl):
     hostname = hostname.split(":")[0]
 
     try:
-        hostname = socket.gethostbyname(hostname)
-    except Exception as e:
+        addrinfo = socket.getaddrinfo(hostname, None)
+    except socket.gaierror:
         # Name resolution failed. This means
         # that the base URL is broken.
         raise BrokenBaseUrl() from e
 
-    country = gi.country(hostname).country.iso_code
-    if not country:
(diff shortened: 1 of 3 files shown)


ott:
This pull request is a first step to proper IPv6 support and support for IPv6-only mirrors.

ott:
IPv6 support is tracked in issue  #306 for the moment.

abompard:
Thanks for updating it! I've fixed a couple remaining issues due to the rebase, and refactored the common code.

abompard:
The world map is generated from the mirror coordinates in the database already. Or did I misunderstood your question?

abompard:
Ah, you may have been tricked into believing that the `generate_worldmap.py` script actually generates the worldmap. It would be too easy! It only stores the host coordinates in the database. The map itself is generated dynamically using the data from the `mirrors_location` view in `views.py`. The unfortunate name of this module is historic: it used to generate a static worldmap (in image format). Now it only stores the coordinates in the database.

PR #312, opened 2022-07-15
Packit

LenkaSeg:
Add `.packit.yml` file.
Make changes in the spec file necessary for rpmlint to pass without errors.
Sync spec file with downstream.
Get rid of the downstream patch.
Increase the version to 0.17

### DIFF
--- a/mirrorlist/mirrorlist_client.wsgi
+++ b/mirrorlist/mirrorlist_client.wsgi
@@ -1,5 +1,3 @@
-#!/usr/bin/env python3
-#
 # Copyright (c) 2014 Red Hat, Inc.
 # Copyright (c) 2007-2013 Dell, Inc.
 #  by Matt Domsch <Matt_Domsch@dell.com>
--- a/mirrorlist/mirrorlist_client.wsgi
+++ b/mirrorlist/mirrorlist_client.wsgi
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 #
 # Copyright (c) 2014 Red Hat, Inc.
 # Copyright (c) 2007-2013 Dell, Inc.
--- a/utility/mirrormanager2.spec
+++ b/utility/mirrormanager2.spec
@@ -1,7 +1,3 @@
-# This package depends on automagic byte compilation
-# https://fedoraproject.org/wiki/Changes/No_more_automagic_Python_bytecompilation_phase_2
-%global _python_bytecompile_extra 1
-
 %if (0%{?rhel} && 0%{?rhel} <= 7)
 # Since the Python 3 stack in EPEL is missing too many dependencies,
 # we're sticking with Python 2 there for now.
@@ -14,7 +10,7 @@
 %endif
 
 Name:           mirrormanager2
-Version:        0.16
+Version:        0.17
 Release:        1%{?dist}
 Summary:        Mirror management application
 
@@ -23,14 +19,13 @@ Summary:        Mirror management application
 # to generate the worldmaps are licensed under GPLv2 and GPLv2+
 License:        MIT and GPLv2+ and GPLv2
 URL:            https://github.com/fedora-infra/mirrormanager2/
-Source0:        %{url}/archive/%{version}/%{name}-%{version}.tar.gz
+Source0:        mirrormanager2-0.17.tar.gz
 
 BuildArch:      noarch
 
 BuildRequires:  python%{python_pkgversion}-devel
 BuildRequires:  python%{python_pkgversion}-flask
 BuildRequires:  python%{python_pkgversion}-flask-admin
-BuildRequires:  python%{python_pkgversion}-flask-oidc
 BuildRequires:  python%{python_pkgversion}-flask-xml-rpc
 BuildRequires:  python%{python_pkgversion}-flask-wtf
 BuildRequires:  python%{python_pkgversion}-wtforms
@@ -70,7 +65,6 @@ BuildRequires:  python%{python_pkgversion}-coverage
 
 Requires:  python%{python_pkgversion}-flask
 Requires:  python%{python_pkgversion}-flask-admin
-Requires:  python%{python_pkgversion}-flask-oidc
 Requires:  python%{python_pkgversion}-flask-xml-rpc
 Requires:  python%{python_pkgversion}-flask-wtf
 Requires:  python%{python_pkgversion}-wtforms
@@ -159,6 +153,7 @@ BuildArch:      noarch
 
 Requires:  %{name}-filesystem = %{version}-%{release}
 Requires:  %{name}-lib = %{version}-%{release}
+Requires:  logrotate
 Requires(pre):  shadow-utils
 
 %description backend
@@ -187,6 +182,7 @@ BuildArch:      noarch
 
 Requires:  %{name}-filesystem = %{version}-%{release}
 Requires:  %{name}-lib = %{version}-%{release}
+Requires:  logrotate
 Requires:  python%{python_pkgversion}-geoip2
 Requires:  python%{python_pkgversion}-matplotlib
 Requires:  python%{python_pkgversion}-basemap
@@ -208,7 +204,7 @@ Base directories used by multiple subpackages
 
 
 %prep
-%setup -q
+%setup -q -n mirrormanager2-0.17
 
 
 %build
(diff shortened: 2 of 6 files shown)


mirrorlist/mirrorlist_client.wsgi
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
adrianreber, line 1: This seems problematic as we still deploy on python2 systems.

utility/mirrormanager2.spec
```diff
@@ -23,7 +19,7 @@ Summary:        Mirror management application
 # to generate the worldmaps are licensed under GPLv2 and GPLv2+
 License:        MIT and GPLv2+ and GPLv2
 URL:            https://github.com/fedora-infra/mirrormanager2/
-Source0:        %{url}/archive/%{version}/%{name}-%{version}.tar.gz
+Source0:        mirrormanager2-0.17.tar.gz
```
adrianreber, line 22: Hardcoding the version seems unnecessary.

utility/mirrormanager2.spec
```diff
@@ -208,7 +206,7 @@ Base directories used by multiple subpackages
 
 
 %prep
-%setup -q
+%setup -q -n mirrormanager2-0.17
```
adrianreber, line 209: This also seems unnecessary.

utility/mirrormanager2.spec
```diff
@@ -255,28 +251,28 @@ install -m 644 utility/mirrormanager2.cfg.sample \
 
 # Install crawler logrotate definition
 install -m 644 utility/mm2_crawler.logrotate \
-    $RPM_BUILD_ROOT/%{_sysconfdir}/logrotate.d/mm2_crawler
+    $RPM_BUILD_ROOT/%{_sysconfdir}/logrotate.d/mirrormanager2-crawler
```
adrianreber, line 254: All the renames from mm2- to mirrormanager2- are also unexpected.

utility/mirrormanager2.spec
```diff
@@ -255,28 +251,28 @@ install -m 644 utility/mirrormanager2.cfg.sample \
...
-install -m 644 utility/mm2_umdl.logrotate \
-    $RPM_BUILD_ROOT/%{_sysconfdir}/logrotate.d/mm2_umdl
+install -m 755 utility/mm2_umdl.logrotate \
+    $RPM_BUILD_ROOT/%{_sysconfdir}/logrotate.d/mirrormanager2-backend
 
 # Install WSGI file
 install -m 644 utility/mirrormanager2.wsgi \
     $RPM_BUILD_ROOT/%{_datadir}/mirrormanager2/mirrormanager2.wsgi
-install -m 644 mirrorlist/mirrorlist_client.wsgi \
+install -m 755 mirrorlist/mirrorlist_client.wsgi \
```
adrianreber, line 263: Not sure 755 is necessary for a wsgi script.

utility/mirrormanager2.spec
```diff
@@ -401,8 +398,7 @@ MM2_SKIP_NETWORK_TESTS=1 ./runtests.sh -v
 
 %files mirrorlist
 %config(noreplace) %{_sysconfdir}/httpd/conf.d/mirrorlist-server.conf
-%attr(755,mirrormanager,mirrormanager) %dir %{_localstatedir}/run/mirrormanager
-%attr(755,mirrormanager,mirrormanager) %dir %{_localstatedir}/lib/mirrormanager/
+%attr(755,root,root) %dir %{_localstatedir}/lib/mirrormanager/
```
adrianreber, line 401: This are also unexpected changes.

utility/mirrormanager2.spec
```diff
@@ -576,6 +643,9 @@ MM2_SKIP_NETWORK_TESTS=1 ./runtests.sh -v
 - umdl: various fixes found in production
   https://github.com/fedora-infra/mirrormanager2/pull/211
 
+* Wed Jul 26 2017 Fedora Release Engineering <releng@fedoraproject.org> - 0.8-2
+- Rebuilt for https://fedoraproject.org/wiki/Fedora_27_Mass_Rebuild
+
```
adrianreber, line 648: Not sure we need to copy all of the changelog entries.

adrianreber (changes requested):

mirrorlist/mirrorlist_client.wsgi
LenkaSeg: Ok. It was a result of this error I was getting at `fedpkg mockbuild`:
```
*** ERROR: ambiguous python shebang in /usr/share/mirrormanager2/mirrorlist_client.wsgi: #!/usr/bin/env python. Change it to python3 (or python2) explicitly.
```
Changed it to `#!/usr/bin/env/python2` and it's passing.

utility/mirrormanager2.spec
LenkaSeg: Mmm...I think it might have been initiative of packit...changing back

utility/mirrormanager2.spec
LenkaSeg: True, the name is the same, removing. I wonder what the `-q` option means,  I don't see it anywhere in the documentation: http://ftp.rpm.org/max-rpm/s1-rpm-inside-macros.html

utility/mirrormanager2.spec
LenkaSeg: The renaming was to fix the rpmlint error `incoherent-logrotate-file`. Seems that rpmlint insist on the logrotate file to be of the same name as the package name: https://fedoraproject.org/wiki/Common_Rpmlint_issues#incoherent-logrotate-file

utility/mirrormanager2.spec
LenkaSeg: This was another error from rpmlint, I think it was `non-executable-script` Changing the permissions to 755 solved it.

utility/mirrormanager2.spec
LenkaSeg: This was also rpmlint error.
Line 404 was `dir-or-file-in-var-run`. Here similar bugzilla report: https://bugzilla.redhat.com/show_bug.cgi?id=1438478
Line 405 was throwing some error too, but I'm unable to reproduce it anymore. Well, ok, if it does not complain now I can change it back to `mirrormanager` instead of `root` :)

utility/mirrormanager2.spec
LenkaSeg: This is a result of updating the upstream spec file with the downstream changes. From now on, packit should propose a PR for downstream from the updated spec file in upstream. In case I don't include all of these Changelog entries from downstream, they might be lost.

utility/mirrormanager2.spec
adrianreber: I think `-q` is to not print out the output of untarring.

utility/mirrormanager2.spec
adrianreber: But this will break existing installations. We should not do that.

utility/mirrormanager2.spec
adrianreber: This is never directly executed, so we should keep it at 644.

utility/mirrormanager2.spec
adrianreber: From my point of view it does not make much sense to carry downstream commit messages in the upstream spec file.

I am undecided. I probably would just go with `%autochangelog` but that does not work on EPEL 7. I think. I would skip this change.

mirrorlist/mirrorlist_client.wsgi
adrianreber: That is also incorrect. It should be python2 on Python 2 system and python3 on other systems.

LenkaSeg: Is the shebang really necessary here? Python2 is unsupported.

utility/mirrormanager2.spec
LenkaSeg: Moving `incoherent-logrotate-file` error to rpmlintrc.

utility/mirrormanager2.spec
LenkaSeg: Removing downstream commit messages.

utility/mirrormanager2.spec
LenkaSeg: Adding `non-executable-script` error to rpmlintrc file.

mirrorlist/mirrorlist_client.wsgi
LenkaSeg: Shebang found redundant and removed.

adrianreber: Good idea.

LenkaSeg:
`ci / ci (registry.fedoraproject.org/fedora:rawhide) (pull_request) ` failing for reasons unrelated to this PR. Currently flask-oidc is undergoing major upgrade.

PR #320, opened 2023-08-10
Modernize the codebase

abompard:
This is a gigantic PR. The idea is to modernize the codebase throughout, for example:
- drop python2 support
- drop obsolete stuff
- adjust to changes in APIs, 
- switch the authentication system to OIDC
- refactor some big modules into separate modules
- use pytest for the tests
- etc...

Hopefully the commits make sense and are topical.


abompard:
OK, the two CI errors are expected, and the F39 one should start passing when the [Bodhi update for python-flask-oidc](https://bodhi.fedoraproject.org/updates/FEDORA-2023-eb7d8ab878) is pushed to stable. The F38 will keep failing though, because it's a stable release and I won't update flask-oidc there.

PR #337, opened 2024-03-20
Require SQLAlchemy < 2

nphilipp:


nphilipp:
Actually, this doesn’t look like it’s necessary.

PR #345, opened 2024-06-04
Fix crawler

james58899:
This fixes the crawler incorrectly marking directories as out of date.


abompard (approved):
Thanks! I'll deploy the changes asap

PR #361, opened 2024-07-10
Bump UA to match the version

cicku:


abompard:
Good idea! I've done it in 64650486 to set the version dynamically, so we don't forget again :-)

PR #391, opened 2024-09-16
Pagination

ryanlerch:


mirrormanager2/templates/fedora/_macros.html
```diff
@@ -280,3 +280,34 @@ <h4>Export Compliance</h4>
...
+                <a class="page-link" href="{{ result.page_url(page_number) }}">
+                  {{ page_number }}
+                </a>
+              </li>
+            {% endif %}
+          {% endfor %}
+        </ul>
+      </nav>
+    {% endif %}
+  {% endmacro %}
```
abompard, line 313: I think the indentation is wrong on this one, the `macro` and the `if` were on the same level at the start.

ryanlerch: fixed this issue here, but that whole file is kinda indented wacky to begin with -- i will fix this in a differeent PR

ryanlerch:
All issues should be resolved now and ready for re-review.

abompard (commented):
Looks good! Just one small thing that looks unrelated.

tox.ini
```diff
@@ -40,7 +40,7 @@ commands =
 
 [testenv:lint]
 commands =
-    poetry run ruff check .
+    poetry run ruff check . --fix
```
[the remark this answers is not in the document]
abompard: If you want to be able to add `--fix` when running tox, you can use `{posargs}`.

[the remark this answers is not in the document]
ryanlerch: yeah ill pull that out -- ive had issues with using the posargs before for some reason -- so i usually do it this way out of habit

mirrormanager2/views.py
```diff
@@ -70,6 +71,17 @@ def list_mirrors(p_name=None, p_version=None, p_arch=None):
...
+    try:
+        page_number = int(flask.request.args.get("page_number"))
+    except (TypeError, ValueError):
+        page_number = 1
+
+    try:
+        page_size = int(flask.request.args.get("page_size", default_page_size))
+    except (TypeError, ValueError):
+        page_size = default_page_size
+
```
abompard, line 84: If you ever want to reuse the `PaginationArgs` dataclass in another view, you'll have to repeat this conversion block. So I would put that in the constructor of the `PaginationArgs` class (I think you can do something like that with dataclasses). The aim would be to just pass it `flask.request.args.get("page_number")` and `flask.request.args.get("page_size")` and let it do the conversion.
Since it's already a flask-specific class, we could even go as far as letting it set those attributes itself, by having it inspect `flask.request`. This way we'll have less repetition of code without sacrificing the abstraction layer! :-)

abompard (commented):
Looking good! Just a possible improvement, nothing mandatory.

mirrormanager2/views.py
ryanlerch: Thanks yeah i thought about this but didnt investigate if it was possible with dataclasses -- looked at using __post_init__ in dataclasses, but since we are really just setting ti all ourselves in the dataclass itself, i just defined my own __init__ in the dataclass with the logic that previously was in the veiw endpoint.

also added a {posargs} to the lint env in tox so i can `-- --fix`

abompard (approved):
Perfect, thanks!

Issue #352, opened 2024-06-21
Add flask-healthz


abompard:
We should add flask-healthz to enable liveness and readiness checks in openshift.

rod7760:
I'll take a look at this!

PR #398, opened 2024-10-03
Added health checks with flask-healthz

rod7760:
Resolves: #352

Signed-off-by: Riccardi Dalexis <rickdalexisdev@gmail.com>


rod7760:
I resolved the failing lint checks. My apologies.

mirrormanager2/health_checks.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def liveness():
+    pass
+
+
+def readiness():
+    try:
+        DB.manager.get_status()
+    except OperationalError as err:
```
abompard, line 14: This looks good, but we can do even better! The sqlalchemy-helpers doc has an example implementation of a health check: https://sqlalchemy-helpers.readthedocs.io/en/latest/user.html#openshift-health-checks

abompard (changes requested):
Thanks for your contribution! There's a possible improvement that I noted, but it's overall very good!

mirrormanager2/health_checks.py
rod7760: Ah. I'm unsure how I missed that, but PR has been updated. I also realized I could simplify the tests.

Thanks

abompard (approved):
Looks great, thanks again!

Issue #472, opened 2025-04-15
[Feature request] Search filters for Mirror Manager


triatic:
Aside from being slow to load, there are currently no options on the Mirror Manager to filter by country: https://mirrormanager.fedoraproject.org/mirrors/Fedora/42

This means for someone to find a mirror in their country they have to click on a page number, wait for a slow page to load, only to find they are at "F" in the alphabet instead of "M".

Can country search filters be added please?

sdglitched:
@abompard, is this issue up for grab? I would love to give it a try.

I believe having a new endpoint something like `@views.route("/mirrors/<p_name>/<p_version>/<p_country>)` can be made available where p_country can be used to filter the list.

Do you think this is the right way forward?

abompard:
Hey! Sorry I was on vacation, I only see your comment now.
I think it would be preferrable to have a query-string-based search filter than to have a path-based one, because I would like to be able to list all mirrors in, for example, US, without choosing a version. So, something like `/mirrors/Fedora/?country=US`.

sdglitched:
No worries! Hope you had nice vacation.

Oh. That's a good take, it'll sure make us apply the filter on any `mirrors` endpoint. It was also bothering me that I had to have all the paths present before getting to filter based on country.

I'll make the changes in the raised PR.

PR #506, opened 2025-07-30
Add country filter with contextual filter display

sdglitched:
Added a new filter parameter to filter `mirrors` based on country code with contextual filter display.

Changes:
- Extended mirrors endpoint to support additional parameter for country code <p_country>
- Added country_id parameter to `get_mirrors()` function to filter mirrors by joining `HostCountry` and `Country` tables
- Replaced static filter text with dynamic `valid_filters` array that builds descriptive text based on active filters which is then displayed with contextual filter descriptions

The implementation allows users to filter mirrors by country while maintaining backward compatibility with existing URLs. Filter descriptions are dynamically generated based on which parameters are present.

Fixes: #472


sdglitched:
@abompard, I've made the filter to work as a query-based parameter rather a separate endpoint.

Thank you for reviewing the PR and let me know if something else needs to be changed/added.

sdglitched:
@abompard, dropping a quick reminder to keep it on radar.

mirrormanager2/templates/fedora/mirrors.html
```diff
@@ -8,11 +8,14 @@
...
-  {{ request.view_args.p_version }}
-  {% if request.view_args.p_arch %}on {{ request.view_args.p_arch }}{% endif %}
+  {% set filter_text = valid_filters | join(' ') %}
+  {% if filter_text.startswith('on ') or filter_text.startswith('in ') %}
+    These are the mirrors {{ filter_text }}
+  {% else %}
+    These are the mirrors for {{ filter_text }}
+  {% endif %}
 </p>
 {% endif %}
```
abompard, line 20: This way of building the sentence using prefixes seems a little bit brittle.
Also, it's a line that could be merged with the following, having both seems redundant.

I suggest making this change to `views.py`:

```diff
diff --git i/mirrormanager2/views.py w/mirrormanager2/views.py
index 85b171c..f9ce427 100644
--- i/mirrormanager2/views.py
+++ w/mirrormanager2/views.py
@@ -62,30 +62,35 @@ def list_mirrors(p_name=None, p_version=None, p_arch=None):
     arch_id = None
     product_id = None
     country_id = None
-    valid_filters = []
+    filters: dict[str, str | int | None] = {
+        "product": None,
+        "version": None,
+        "arch": None,
+        "country": None,
+    }
     if p_name:
         product = mmlib.get_product_by_name(DB.session, p_name)
         if product:
             product_id = product.id
-            valid_filters.append(p_name)
+            filters["product"] = p_name
             if p_version:
                 version = mmlib.get_version_by_name_version(DB.session, p_name, p_version)
                 if version:
                     version_id = version.id
-                    valid_filters.append(p_version)
+                    filters["version"] = p_version
 
     if p_arch:
         arch = mmlib.get_arch_by_name(DB.session, p_arch)
         if arch:
             arch_id = arch.id
-            valid_filters.append(f"on {p_arch}")
+            filters["arch"] = p_arch
 
     p_country = flask.request.args.get("country")
     if p_country:
         country = mmlib.get_country_by_name(DB.session, p_country)
         if country:
             country_id = country.id
-            valid_filters.append(f"in {p_country}")
+            filters["country"] = p_country
 
     mirrors = mmlib.get_mirrors(
         DB.session,
@@ -111,7 +116,7 @@ def list_mirrors(p_name=None, p_version=None, p_arch=None):
     return flask.render_template(
         "mirrors.html",
         mirrors=mirrors,
-        valid_filters=valid_filters,
+        filters=filters,
         countries=countries,
         selected_country=p_country,
     )
```

It will build a `filters` dictionary instead of a list, and we can use the key to know how to forge the sentence. In the template:

```diff
-{% if valid_filters and valid_filters|length > 0 %}
-<p>
-  {% set filter_text = valid_filters | join(' ') %}
-  {% if filter_text.startswith('on ') or filter_text.startswith('in ') %}
-    These are the mirrors {{ filter_text }}
-  {% else %}
-    These are the mirrors for {{ filter_text }}
-  {% endif %}
-</p>
-{% endif %}
-
 {% if mirrors %}
 <p>
   We have currently {{ mirrors | length }} active mirrors
+  {% if filters["product"] %} for {{filters["product"]}} {% endif %}
+  {% if filters["version"] %} {{filters["version"]}} {% endif %}
+  {% if filters["arch"] %} on {{filters["arch"]}} {% endif %}
+  {% if filters["country"] %} in {{filters["country"]}} {% endif %}
 </p>
```

What do you think?

mirrormanager2/templates/fedora/mirrors.html
```diff
@@ -21,6 +24,35 @@ <h2>Public active mirrors</h2>
...
+    <select class="form-select" id="country-filter" name="country" style="width: auto;">
+      <option value="">All</option>
+      {% for country in countries %}
+        <option value="{{ country.code }}" {% if selected_country == country.code %}selected{% endif %}>
+          {{ country.code }}
+        </option>
+      {% endfor %}
+    </select>
+  </div>
+</div>
```
abompard, line 39: I think we can move this filter inside the column header:

```diff
 <table class="table table-striped">
   <tr>
-    <th>Country</th>
+    <th>
+    <i class="bi bi-globe fs-4"></i>
+      <select class="form-select form-select-sm d-inline w-auto ms-1" id="country-filter" name="country">
+        <option value="">All</option>
+        {% for country in countries %}
+          <option value="{{ country.code }}" {% if selected_country == country.code %}selected{% endif %}>
+            {{ country.code }}
+          </option>
+        {% endfor %}
+      </select>
+    </th>
     <th>Site Name</th>
     <th>Mirror Name</th>
```

If that's fine with you, please also move the `script` tag below the table.

abompard (changes requested):
A couple suggestions, thanks!

mirrormanager2/templates/fedora/mirrors.html
sdglitched: I also thought about it when I started implementing the filter, but the out of sync headers made me rethink.

I tried your implementation and still feel the same about it.
image

Yet this seems to be the way as it will not have a weird single filter hanging alone and taking up space.

mirrormanager2/templates/fedora/mirrors.html
sdglitched: This indeed is a more cleaner approach. I'll incorporate it.

mirrormanager2/templates/fedora/mirrors.html
abompard: Ah I see! That's why I proposed just the icon as the header and then, on the side of it, the dropdown. I think it looks less "different"

mirrormanager2/templates/fedora/mirrors.html
```diff
@@ -8,22 +8,27 @@
...
+  {% if filters["version"] %} {{filters["version"]}} {% endif %}
+  {% if filters["arch"] %} on {{filters["arch"]}} {% endif %}
+  {% if filters["country"] %} in {{filters["country"]}} {% endif %}
 </p>
 
 <table class="table table-striped">
   <tr>
-    <th>Country</th>
+    <th>
+      <select class="form-select form-select-sm d-inline w-auto ms-1" id="country-filter" name="country">
```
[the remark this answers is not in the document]
abompard: Actually, don't worry about it, I'll fix it afterwards, it's so very minor :-)

abompard (approved):

PR #519, opened 2025-09-05
Add REST API `checkin` endpoint alongside XML-RPC

### LINKED ISSUE #511: Replace the XML-RPC operations in the `report_mirror` client and the server by a REST API.
The Flask XML-RPC library dependency is unmaintained, has been forked, and the forked is unmaintained as well.
There's a single operation that only takes a file as argument, it could easily be replaced by a "REST" (rather, HTTP) API with an single POST-only endpoint.

sdglitched:
This PR modernises the mirror `checkin` system by replacing the `XML-RPC` interface with a `REST API` endpoint.

Key Changes:
- Add REST API checkin endpoint alongside XML-RPC

@abompard, I've also dropped the pickle support as you suggested. Thank you for reviewing and let me know if any changes are required.

Fixes: #511


sdglitched:
@abompard: Sorry, I completely missed your comment and forgot about this PR.

I've made the change so that existing XML-RPC clients continue to work while new clients migrate to REST.

Once you are available, please review this PR.

client/report_mirror
```diff
@@ -7,10 +7,12 @@ import copy
 import json
 import os
 import pprint
+import requests
 import sys
-import xmlrpc.client as xmlrpclib
 from argparse import ArgumentParser
 
+from flask import request
+
```
abompard, line 14: It does not look like you are using this. Which is a good thing because this is a client-side script, you can't expect Flask to be installed on the client :-)
You can remove this line.

client/report_mirror.conf
```diff
@@ -1,8 +1,8 @@
 [global]
 # if enabled=0, no data is sent to the database
 enabled=1
-# server= is the URL to the MirrorManager XML-RPC interface
-server=https://mirrormanager.fedoraproject.org/xmlrpc
+# server= is the URL to the MirrorManager REST interface
```
abompard, line 4: The blueprint in `api.py` is registered with `url_prefix="/api"` in `app.py`, so the actual endpoint URL is `/api/checkin`, not `/checkin`.

mirrormanager2/api.py
```diff
@@ -151,3 +157,30 @@ def api_repositories():
...
+
+        uncompressed = bz2.decompress(base64.urlsafe_b64decode(data["data"]))
+        config = json.loads(uncompressed)
+        r, host, message = read_host_config(DB.session, config)
+
+        if r is not None:
+            logging.info(f"Checkin for host {host} successful: {message}")
+            return (
+                flask.jsonify({"success": True, "message": message + "checked in successful"}),
+                200,
```
abompard: Missing space before `"checked in successful"` — this will produce messages like `"Hostname XYZchecked in successful"`. Same issue on line 182 with `"error checking in"`.
I would recommend doing something like `f"{message} checked in successful"` instead.

client/report_mirror
```diff
@@ -283,27 +285,19 @@ def main():
...
-                    "Error checking in.  Connection closed before "
-                    "checkin complete.  Please try again later."
-                )
-                sys.exit(1)
+            response = requests.post(
+                item.config["global"]["server"],
+                json={"data": data},
+                headers={"Content-Type": "application/json"},
+            )
+            if response.status_code == 200:
```
abompard: You can use [response.ok](https://docs.python-requests.org/en/latest/api/#requests.Response.ok) instead, it'll handle all the return codes in the 2xx and 3xx range.

```suggestion
            if response.ok:
```

mirrormanager2/api.py
```diff
@@ -151,3 +157,30 @@ def api_repositories():
...
+            )
+        else:
+            logging.error(f"Error for host {host} during checkin: {message}")
+            return flask.jsonify({"success": False, "error": message + "error checking in"}), 400
+    except Exception as e:
+        logging.error(f"Checkin error: {e}")
+        return flask.jsonify({"success": False, "error": f"Internal server error: {str(e)}"}), 500
```
abompard: The `str(e)` inside a formatted string is not necessary, objects are automatically converted to strings when used in formatted strings.

mirrormanager2/api.py
```diff
@@ -151,3 +157,30 @@ def api_repositories():
...
+
+
+@views.route("/checkin", methods=["POST"])
+def checkin():
+    """REST API endpoint to replace XML-RPC checkin functionality."""
+    try:
+        data = flask.request.get_json()
+
+        if not data or "data" not in data:
+            return flask.jsonify({"error": "Missing data parameter"}), 400
```
abompard: This format is not consistent with the other returned formats, please switch the `error` key to `message`. It will make parsing the result easier on the client side if needed.

As for the others, the `success` boolean is a bit redundant with the HTTP return code, I don't think it's necessary.

abompard (changes requested):
Thanks! Could you also please add some unit tests for the new `checkin` endpoint?

client/report_mirror
```diff
@@ -283,27 +285,19 @@ def main():
...
-                print(
-                    "Error checking in.  Connection closed before "
-                    "checkin complete.  Please try again later."
-                )
-                sys.exit(1)
+            response = requests.post(
+                item.config["global"]["server"],
+                json={"data": data},
+                headers={"Content-Type": "application/json"},
+            )
```
abompard, line 294: Oh yeah one more thing, please add some error handling here as well (as there was before with xml-rpc). You can look at the [RequestException](https://docs.python-requests.org/en/latest/api/#requests.RequestException) class tree.

client/report_mirror
```diff
@@ -7,10 +7,12 @@ import copy
 import json
 import os
 import pprint
+import requests
```
abompard: This is a non-stdlib module, please add in the documentation that it is required.

abompard (changes requested):

client/report_mirror
sdglitched: Will it be ok if I add this as a dependency in the `/client/README`? I'm not able to find any other suitable place for it.

client/report_mirror
sdglitched: Oh, didn't know that :hushed:

client/report_mirror.conf
sdglitched: Uhh.. missed it :crying_cat_face:

client/report_mirror
sdglitched: I had it when I started with the PR, but later forgot to remove it, and the formater also did not remove it :disappointed_relieved:.

client/report_mirror
abompard: Yes!

mirrormanager2/api.py
```diff
@@ -151,3 +157,30 @@ def api_repositories():
...
+        if r is not None:
+            logging.info(f"Checkin for host {host} successful: {message}")
+            return (
+                flask.jsonify({"message": f"{message} checked in successfully"}),
+                200,
+            )
+        else:
+            logging.error(f"Error for host {host} during checkin: {message}")
+            return flask.jsonify({"message": f"{message} error checking in"}), 400
+    except Exception as e:
```
abompard, line 188: `except Exception` catches everything, including `binascii.Error` (bad base64), `OSError` (bad bz2 data), and `json.JSONDecodeError` — all caused by malformed client input. These should return 400, not 500. Consider catching them explicitly before the broad handler:
```python
except (binascii.Error, OSError, json.JSONDecodeError) as e:
    return flask.jsonify({"message": f"Invalid data: {e}"}), 400
except Exception as e:
    ...
```

client/report_mirror
```diff
@@ -283,26 +284,23 @@ def main():
...
-                )
-                sys.exit(1)
-            except xmlrpclib.Fault:
-                print(
-                    "Error checking in.  Connection closed before "
-                    "checkin complete.  Please try again later."
+                response = requests.post(
+                    item.config["global"]["server"],
+                    json={"data": data},
+                    headers={"Content-Type": "application/json"},
```
abompard: The `Content-Type: application/json` header is redundant here. The `requests` library sets it automatically when the `json=` keyword argument is used. This line can be removed.

mirrormanager2/api.py
```diff
@@ -151,3 +157,30 @@ def api_repositories():
...
+
+        uncompressed = bz2.decompress(base64.urlsafe_b64decode(data["data"]))
+        config = json.loads(uncompressed)
+        r, host, message = read_host_config(DB.session, config)
+
+        if r is not None:
+            logging.info(f"Checkin for host {host} successful: {message}")
+            return (
+                flask.jsonify({"message": f"{message} checked in successfully"}),
+                200,
```
abompard: You don't need to specify `200`, it's the default value when only `flask.jsonify()` is returned.

mirrormanager2/api.py
```diff
@@ -151,3 +157,30 @@ def api_repositories():
...
+    """REST API endpoint to replace XML-RPC checkin functionality."""
+    try:
+        data = flask.request.get_json()
+
+        if not data or "data" not in data:
+            return flask.jsonify({"message": "Missing data parameter"}), 400
+
+        uncompressed = bz2.decompress(base64.urlsafe_b64decode(data["data"]))
+        config = json.loads(uncompressed)
+        r, host, message = read_host_config(DB.session, config)
```
abompard, line 174: The `message` strings returned by `read_host_config` (and `validate_config`) end with `\n`. The endpoint then returns `f"{message} error checking in"`, which embeds that newline in the middle of a JSON string value: ugly and unexpected for API consumers. Consider calling `.strip()` on `message` before including it in the response (ideally once, after this line).

abompard (changes requested):
A few things still need attention.

The PR description says *"Removed XML-RPC implementation"* but it is still kept for backward compatibility. Please update the description to match what the code actually does.

There is no test covering wrong site name or password. Since `read_host_config` is the only authentication gate for this unauthenticated endpoint, that path really should be tested. Likewise, there is no test for a host that exists but has `private=False` (which `read_host_config` rejects with a specific message).

mirrormanager2/api.py
sdglitched: I was kinda following the existing pattern, but I understand why specifying it isn't necessary.
https://github.com/fedora-infra/mirrormanager2/blob/953b79fc28f054fe19c6d279a8ba1632c446bf8f/mirrormanager2/api.py#L91-L93

sdglitched:
@abompard, I've made the suggested changes. Please take a look once you are available.

tests/test_checkin.py
```diff
@@ -0,0 +1,214 @@
...
+    compressed = bz2.compress(json.dumps(config).encode())
+    encoded = base64.urlsafe_b64encode(compressed).decode()
+    output = client.post(
+        "/api/checkin",
+        json={"data": encoded},
+    )
+    assert output.status_code == 200
+    data = output.get_json()
+    assert "message" in data
+    assert "checked in successfully" in data["message"]
```
abompard: This test is named `test_checkin_success_no_newlines` and its docstring says *"Test /api/checkin success response has no newlines in message"*, but there is no assertion that verifies the absence of newlines. The assertions here are identical to those in `test_checkin_success`. You should at least add the missing assertion:
```python
assert "\n" not in data["message"]
```
but if you look at the content of `data["message"]`, you'll see that is starts with a space which is a little weird. Please find a way to avoid that that limits duplication of code.

client/report_mirror
```diff
@@ -283,26 +284,22 @@ def main():
 
     if not args.no_send:
         #  print("Connecting to %s" % item.config['global']['server'])
-        server = xmlrpclib.ServerProxy(item.config["global"]["server"])
 
         data = base64.urlsafe_b64encode(bz2.compress(p.encode())).decode()
 
         if data is not None:
```
abompard: `base64.urlsafe_b64encode()` always returns a `bytes` object; it never returns `None`. This `if data is not None:` guard is always `True` and can be removed. (Pre-existing issue, but this PR touches the surrounding code.)

abompard (changes requested):
One more minor thing: the PR description is still inaccurate, it still says "Removed XML-RPC implementation".

tests/test_checkin.py
sdglitched: I forgot to add that assert statement. I'll add it. But I'm not sure which duplication you are mentioning. Is it the assertions in the test cases?

Also, I feel we will need to use some conditional statement to check if the `message` is empty and then generate the response.
```python
msg = message.strip()
result = f"{msg} checked in successfully" if msg else "checked in successfully"
return flask.jsonify({"message": result})
```

sdglitched:
@abompard, it's up for review.

sdglitched:
@abompard, just dropping a quick reminder.

PR #541, opened 2025-11-06
Optimize netblocks processing: 250min → 8min

adrianreber:
Major performance improvements to mirrormanager2/utility/netblocks.py:

- **30x faster execution**: Processing time reduced from 250 minutes to 8 minutes
- **Parallel processing**: Uses two separate processes for IPv4 and IPv6 netblocks, avoiding Python GIL limitations with ProcessPoolExecutor
- **O(n²) → O(1) deduplication**: Replaced list-based duplicate detection with set-based approach, eliminating the primary performance bottleneck
- **Rich progress bars**: Added download progress bars with speed/ETA and real-time BGP processing progress with entry counts
- **Multiprocessing-safe progress**: Implemented queue-based communication between worker processes and main thread for live progress updates
- **Better user experience**: Clean visual feedback during both download and processing phases

The script now efficiently utilizes multiple CPU cores while providing comprehensive progress indication throughout the entire operation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/mirrormanager2/utility/netblocks.py
+++ b/mirrormanager2/utility/netblocks.py
@@ -14,9 +14,13 @@ to an ASN to get them to a mirror that's closer to them than not.
 """
 
 import bz2
+import concurrent.futures
 import logging
+import multiprocessing
 import os
+import queue
 import re
+import threading
 from contextlib import contextmanager
 from datetime import date, timedelta
 from io import BytesIO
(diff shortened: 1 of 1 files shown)


abompard:
Looks nice, thanks! Unfortunately there are currently no unit tests for the netblocks script, so before doing a refactoring of this magnitude I'd prefer to add unit tests with good coverage for the current code and make sure they still pass with the new code. I'll try to write them when I find the time, but feel free to beat me to it if you want to! :-)

adrianreber:
Good point. Let me first introduce some test cases. That should not be too difficult to have a simple test case around this script.

